PR #3, opened 2017-07-31
Put in first pass of colors

lastquestion:
I sat down and showed @mannionaco @lindseyhight how to do this. This is a first pass of colors.

### DIFF
--- a/OTKit/otkit-colors/colors.yml
+++ b/OTKit/otkit-colors/colors.yml
@@ -1,3 +1,15 @@
 props:
-  # TODO
+  primary-action:
+    value: "#F91F14"
   
+  secondary-action:
+    value: "#E81106"
+    
+  primary-gray:
+    value: "#333333"
+    
+  secondary-gray:
+    value: "#717171"
+    
+  utility-gray:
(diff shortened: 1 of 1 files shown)


OTKit/otkit-colors/colors.yml
```diff
@@ -1,3 +1,15 @@
...
+    value: "#E81106"
+    
+  primary-gray:
+    value: "#333333"
+    
+  secondary-gray:
+    value: "#717171"
+    
+  utility-gray:
+    value: "#E1E1E1"
```
lindseyhight: Actually, this is ...

OTKit/otkit-colors/colors.yml
```diff
@@ -1,3 +1,15 @@
...
+  secondary-action:
+    value: "#E81106"
+    
+  primary-gray:
+    value: "#333333"
+    
+  secondary-gray:
+    value: "#717171"
+    
+  utility-gray:
```
lindseyhight: This name is still WIP according @mannionaco

lindseyhight:
We should close this and use #5 instead.

PR #4, opened 2017-07-31
Update spacing.yml

mannionaco:
Just beginning. WIP.

PR #5, opened 2017-08-01
First pass at new colors

lindseyhight:
This replaces #3 - adds comments, fixes up tabbing and adds required props.

Also adds all colors such as secondary, disabled etc.

### DIFF
--- a/OTKit/otkit-colors/colors.yml
+++ b/OTKit/otkit-colors/colors.yml
@@ -1,3 +1,49 @@
 props:
-  # TODO
+
+  # Primary Action
+  # This is only used on objects, backgrounds with large text (24px or larger, bold), and will be used as a color to guide users towards actionable content.
+  # =============================================
   
+  color-primary:
+    value: "#F91F14"
+  
+  color-primary-active:
+    value: "#FB665F"
+  
+  color-primary-disabled:
+    value: "rgba(#F91F14, 0.40)"
+  
+  # Secondary Action
+  # This color is only to be used on text sizes below 24px for actionable content (tabs, links, underlines, points)
+  # =============================================
+  
+  color-secondary:
+    value: "#E81106"
+  
+  color-secondary-active:
+    value: "#FA4940"
+  
+  color-secondary-disabled:
+    value: "rgba(#E81106, 0.40)"
+  
+  # Primary Gray, used on all text sizes 
+  # =============================================
+  
+  color-primary-gray:
+    value: "#333333"
+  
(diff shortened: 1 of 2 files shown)


OTKit/otkit-colors/colors.yml
```diff
@@ -1,3 +1,49 @@
...
+  # =============================================
+  
+  color-primary:
+    value: "#F91F14"
+  
+  color-primary-active:
+    value: "#FB665F"
+  
+  color-primary-disabled:
+    value: "rgba(#F91F14, 0.40)"
```
lindseyhight: @nickbalestra can we even do this? should I use `rgba(0.9,0.1,0.1,1.0)`?

lindseyhight:
I accidentally put spacing in the with colors PR. Let's also close #4

OTKit/otkit-colors/colors.yml
```diff
@@ -1,3 +1,49 @@
...
+  color-secondary-active:
+    value: "#FA4940"
+  
+  color-secondary-disabled:
+    value: "rgba(#E81106, 0.40)"
+  
+  # Primary Gray, used on all text sizes 
+  # =============================================
+  
+  color-primary-gray:
```
francesyun: I think  the gray variable names should be color-gray-primary, color-gray-secondary, color-gray-utility.

This is because color-primary-gray sounds like a subset of color-primary, which it isn't. Also, when you name it color-gray-[x], it's easier to autocomplete on a text editor when you want to include a gray colored text.

OTKit/otkit-spacing/spacing.yml
```diff
@@ -1,3 +1,22 @@
...
+  spacing:
+    value: "1rem"
+  
+  spacing-xsmall:
+    value: "0.25rem"  
+
+  spacing-small:
+    value: "0.5rem"
+  
+  spacing-medium:
```
francesyun: why is there "spacing" and "spacing-medium" if they are the same value?

OTKit/otkit-colors/colors.yml
nickbalestra: Yes you are right, you need to follow the rgba specs for those values. I see that we also have similar error on one specific value on the [OTTheme color token](https://github.com/opentable/design-tokens/blob/master/OTTheme/ottheme-colors/colors.yml#L64), I'll open an issue for this. Good catch 👍

nickbalestra:
Awesome 💅  work @lindseyhight!
Going to merge this now

lastquestion:
💃 💃

PR #9, opened 2017-08-02
[OC-97] OTKit build in cssmodules format

nickbalestra:
- [x] `otkit-*/*.yaml` files need to be renamed into token.yml for the build pipeline to work in a more agnostic way for each token. (DO a PR after #4 and #5 get merged)
- [x] the `cssmodules.css` format for the moment is included until it [gets added in theo](https://github.com/salesforce-ux/theo/pull/111)
`ot-design-token-scripts` to be published on npm on a follow-up pr


lastquestion:
Dope 🎉

PR #13, opened 2017-08-03
Minor adjustements

nickbalestra:
- added minor left out on the various packages in preparation for publishing
- bette defaults for the build so consuming will look like `require('otkit-spacing/token.cssmodules.css')` (no need for a `dist/`)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
 node_modules
 *.log
 .DS_Store
-dist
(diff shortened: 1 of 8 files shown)


lastquestion:
the `dist` thing is super annoying. I wish `npm publish` had some sort of `root` config so we don't have to do build shenanigans.

PR #19, opened 2017-08-04
[OC-104] Output tokens in scss, cssmodules.css and common.js formats

nickbalestra:


nickbalestra:
@matteofigus yes, ditto that., builds are now ignored.

PR #24, opened 2017-08-04
[RUSH-792] Improve documentation of OTKit around save-dev, minor fixes

karuto:
# Description
This PR primarily aims to resolve some confusion around `save-dev` in the current OTKit documentation. It also includes some very minor typo and formatting fixes.

Going forward I plan to continue polishing and adding to the documentations as we develop more with design tokens.

# Ticket
https://opentable.atlassian.net/browse/RUSH-792

### DIFF
--- a/OTKit/README.md
+++ b/OTKit/README.md
@@ -11,15 +11,21 @@
 
 ## Usage
 
-Add the needed tokens to your devDependnecies:
+Run the command below will add the needed tokens to your project:
 
+```bash
+$ npm install --save-dev <token-name>
 ```
-$ npm install save-dev <token-name>
+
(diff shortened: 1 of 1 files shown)


OTKit/README.md
```diff
@@ -11,15 +11,21 @@
 
 ## Usage
 
-Add the needed tokens to your devDependnecies:
+Run the command below will add the needed tokens to your project:
 
-```
+```bash
 $ npm install save-dev <token-name>
```
nickbalestra: I think the correct flag should be `--save-dev` thats why we had that little issue earlier when we tried it.

karuto: You are correct, fixed

PR #28, opened 2017-08-08
[OC-92] Typography token

nickbalestra:
@lastquestion I've created a scaffholding for the typography token.

When working on it, please rely on the [aliases](https://github.com/opentable/design-tokens/blob/master/OTKit/aliases.yml) file for storing all alias/value and then use those within the token itself. 

This allows for non duplicating and reuse of values among tokens if/when needed. (spacing and colors already rely on it). It also make for designer even easier as they most probably want to focus their work on that file first. If you need anything, please do let me know.

### DIFF
--- a/OTKit/otkit-typography/LICENSE
+++ b/OTKit/otkit-typography/LICENSE
@@ -1,0 +1,20 @@
+Copyright (c) 2017 OpenTable, Inc.
+
+Permission is hereby granted, free of charge, to any person obtaining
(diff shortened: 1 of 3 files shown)


nickbalestra:
Thanks! Aliases exist is [here](https://github.com/opentable/design-tokens/blob/master/OTKit/aliases.yml), it comprises all the aliases for all the design system, therefore is in root 😎

Issue #26, opened 2017-08-06
let's think about fonts


lastquestion:
## values for fonts
So, we need to support a variety of values for fonts. We need to transport

* font weights
* font sizes
* line heights
* font family

All of these are fine to transport via values.

So in particular, I think we can be pretty low-level here in design-token land. I would imagine something like this: (values made up, names made up, etc.):

```
regular-weight: 200;
medium-line-height: 1rem;
medium-size: 0.875rem;
regular-font-family: 'source-sans, arial';
```

(One little thing here. I like the idea of not being too verbose, so since we're already in fonts, why do we need to throw font in the names of stuff everywhere? e.g. `medium-size` versus `medium-font-size`.)

## Not encoding @font-face or where the fonts are located
I would like not to deal with `@font-face` here. It's a platform specific problem of web, and doesn't apply to native platforms (which we care about, in the future). Additionally, I'd like to ***not*** encode the way that the font is found in the design token.  

So I will open another issue in react land to deal with that - how to encode `@font-face`, and where the font is actually loaded. There is some interaction with `:global` and CSS-modules (bugs, basically, in webpack's `css-loader`). I think we also need to consider that `@font-face` needs to be a global rule, and cannot be a `local` in a CSS module. 

## how to encode text styles?
From a design perspective, we absolutely have "groupings" of "font design tokens" that define a "text style" (I made these words up just now). But the way to encode this becomes interesting. 

We want to define the "text style name", and the "attribute values for the font style", in a non-platform specific way. Therefore, we should probably encode them like this:

```
text-style-1:
  weight: "{!regular-weight}
  size: "{!medium-size}"
  line-height: "{!medium-line-height}"
  font-family: "{!brand-family}"

text-style-2:
  ....
```

A couple of points here:

* no, those names do not match CSS names, on purpose. platform agnostic, yo.
* Alias probably works ok here, since when we change the alias value, it will change everywhere
* I used `text-style-1` to avoid discussions about `h1`-`h4`. 👼 

The idea is we extend `theo` to generate correct representations of this in each target format. For example, in CSS-module, I'd propose we generate:

```
@value text-style-1: "200 0.875rem/1rem source-sans, arial"
```

(I may have got the syntax slightly off). In other words, that is, provide a singular **value** that later we can import as

```
font: @text-style-1
```

[...]

PR #32, opened 2017-08-11
Adding typography styles token

lastquestion:
<del>1. Create `h1.token.yml` file that expose `size, color, weight, line-height`. 
2. These get generated into full `h1.token.cssmodules.css` (and common.js, and SCSS).
   To do this, I edited the `theo-cli` to support directories.
3.Add a `h1.type.cssmodules.css`, which composes the `h1.token.cssmodules.css` (which is autogenerated) into a `.text-syle` which can then be used like this:
   ```
   .restaurant-title-text {
     composes: textStyle from '....h1.type.cssmodules.css';
   }
   ```
<del>I'm not married to anything here, see related issue #26 for some context & motivation. So, to add a new text-style, you create a new `whatever.token.yml`, and a `whatever.type.cssmodules.css`.</del>

<del>I think long term it would be super interesting to consider have theo understand this kind of groupings or file formats and have it generate a special format. I played around with this and can see how it would be done, but I'm not sure if that's the right direction to explore yet.</del>

Add a new design token NPM module called `otkit-typography-styles`. Each of the styles express 4 values:
* `size`
*  `weight`
* `line-height`
* `color`

All of which can be used by each platform appropriately.

Each style value is kebab cased, so for example `subheading`'s `weight` is `subheading-weight`.

### DIFF
--- a/OTKit/otkit-typography/package.json
+++ b/OTKit/otkit-typography/package.json
@@ -1,6 +1,6 @@
 {
   "name": "otkit-typography",
-  "version": "1.0.1",
+  "version": "1.0.2",
   "description": "OpenTable typography design token",
   "author": {
     "name": "Nick Balestra",


nickbalestra:
I see what you doing here and I can understand your intent.

Is important that we approach tokens in very caution ways and resist yet to the temptation of making those "higher order token" to better exploit some feature of a specific format (in this pr proosal css-modules `compose`). As you pointed out in #26 , such composition will be more appropriate to exist  in the consumer-land where you'll be free to build such classes out of tokens and recompose them, and actually will totally make sense there. The less opinionated tokens will be the more successful their adoption.

Some other thoughts:
- This could makes the token system too convolute and complex on two side: consumer-land: with a more fragmented API and "sub" tokens to imports and here on the producer-land side with a less obvious structure (we are aiming for max contribution herw, therefore we should make an effort to reduce barrier not viceversa). 
- Could drive the token project more toward a specific format.
- The very same result could be achieved at consumer site, where I can totally see you build those groups and classes to be composed, so yes, this feel definitely more at home in a repo like the ot-react-ui and therefore not really needed here.

lastquestion:
I'll write more later, but yes, I want to make sure that code doesn't matter here. It's a good way to start the conversation. When doing this work, I was more interested in starting some thinking then anything else.

lastquestion:
Hm, after some time thinking, here are some of my thoughts about this.

* I think building "the actual implementation of the token" in the point of consumption seems valid. I want to think about that problem because it pushes down consistency and re-implementation issues further down the stack. It's a balance though. I think CSS and SCSS is just "too easy" for people to slip into more and more complicated rules, e.g. implementation. And I think that's what I also fell into too. As an analogy, we'd never export an Java .JAR class to correctly construct fonts in this repo, so why we would do that for CSS?

* But, I believe that "H1 text style", or "H2 on dark background", or "H2 on light background", or "subtext", are themselves tokens, and I want to think about how to encode those tokens. This would not be the implementation of those tokens - but, I would like to encode an abstract grouping of simpler tokens. The way I did this here was to generate additional files that depict the "values" of the "token", in a platform agnostic way. I think this part is good and worth further discussion. I considered using a CSS string format as well, as it's easy to parse by multiple platforms. But really, JSON is the better interchange format here. So I kind of like the way I did it here by building multiple files.

lastquestion:
^- I removed the fancy CSS rule file

lastquestion:
I simplified the problem by making a new NPM module, this removes the Theo CLI changes, and then I just use kebab case to namespace the values into groups. Simple and effective.

This has most of the styles, down to subheading, but it's missing `body` and `subtext` (and variants of that)

OTKit/otkit-typography-styles/token.yml
```diff
@@ -0,0 +1,69 @@
...
+    value: "{!font-weight-bold}"
+  h2-line-height:
+    value: "{!line-height-large}"
+
+  h3-size:
+    value: "{!font-size-medium}"
+  h3-color:
+    value: "{!color-gray-primary}"
+  h3-weight:
+    value: "{!font-weight-normal}"
```
lastquestion: `h3` and `h2` used weight `normal` and `medium` interchangeably in the invision board. I'm pretty sure you meant `normal` for both. cc @mannionaco @francesyun @lindseyhight

nickbalestra: Now that you mention this, I noted you [used](https://github.com/opentable/design-tokens/blob/master/OTKit/aliases.yml#L81) 'normal' and 'bold' what that does it means? I think that they should have an actual weight value I guess @lastquestion ?

karuto: @nickbalestra The font weight "normal" = 400 and "bold" = 700. I agree that these can be changed into actual numerical weight values as they have no feedback. 

The only special gotcha in CSS font weight is the fallback which only applies for numeric 400 and 500 values, so really "normal" and "bold" is the same as 400 and 700 here.

nickbalestra: Yup, I'm trying to think in general terms here and not too tight with CSS :)

lastquestion:
I like the name of headings also. Let's change that.

About having a separate token repo: I thought you wanted that. Let me fix that too. :)

lastquestion:
I'll reopen this with a rebased commit.

PR #38, opened 2017-08-29
changed primary and secondary colors

lindseyhight:
Secondary and primary are currently the same, pending new color from brand in future

### DIFF
--- a/OTKit/aliases.yml
+++ b/OTKit/aliases.yml
@@ -7,21 +7,11 @@ aliases:
   # This color is only used on objects, backgrounds with large text (24px or larger, bold), and will be used as a color to guide users towards actionable content.
   # =============================================
   color-primary:
-    value: "#F91F14"
+    value: "#DA3743"
   color-primary-active:
     value: "#FB665F"
   color-primary-disabled:
-    value: "rgba(249, 31, 20, 0.40)"
-
-  # Secondary Action
-  # This color is only to be used on text sizes below 24px for actionable content (tabs, links, underlines, points)
-  # =============================================
-  color-secondary:
-    value: "#E81106"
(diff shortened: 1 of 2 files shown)


nickbalestra:
@lindseyhight can you please provide some context here? What is the rationale of having primary and secondary the same? Could be a viable option to just remove the secondary until we have one and tag this release as a braking change, and then release a a new minor again once we do have it? (As i don't think people will use a secondary if its the same of a primary, this should help misuse) @lastquestion what you suggest as strategy for handling such releases, and this very specific case?

lindseyhight:
@nickbalestra we already had spec'd some elements in secondary, but I'll just ask the team to update that token everywhere to primary and then we can remove the secondary token until we have one like you suggested.

lastquestion:
@nickbalestra I agree with your feelings. I think it would be good to publish a breaking change version with just `primary`. We'll practice pushing design choice changes through the system via this change, which is also important. 

could you work with @lindseyhight to change this PR to do that?

nickbalestra:
Good idea @lastquestion !
@lindseyhight as we don't have yet an automatic way to publish will be important to state in the PR title that this is a braking change, something like `[BRAKING CHANGE] - your title` could work. This will make clear for who's in charge of publishing that will need to publish that specific token as a major version in this case. Maybe specifying why/what the braking change is in the description could also a good practice.

versions works like this:
```
MAJOR version: when you make incompatible changes,
MINOR version: when you add functionality in a backwards-compatible manner, and
PATCH version: when you make backwards-compatible bug fixes.
```

In this case will be a MAJOR version.

@lindseyhigh Could you try changing this pr to reflect on those suggestion? (feel free to reachout on hipchat if you need any help)

PR #39, opened 2017-09-01
introduced pixel value design tokens for layout of new restaurant profile

ericz82:
When setting `width`, `height` and conditions for media queries, we want to avoid hard-coding the pixel values and instead have them live here as design tokens. Let me know what your thoughts are.

### DIFF
--- a/OTKit/aliases.yml
+++ b/OTKit/aliases.yml
@@ -85,3 +85,9 @@ aliases:
     value: "bold"
   font-name:
     value: "Brandon Text"
+
+  # #############################################
+  # Layout
+  # #############################################
+  breakpoint-desktop-tablet:
+    value: 1056
--- a/OTKit/otkit-breakpoints/LICENSE
+++ b/OTKit/otkit-breakpoints/LICENSE
@@ -1,0 +1,20 @@
+Copyright (c) 2017 OpenTable, Inc.
+
+Permission is hereby granted, free of charge, to any person obtaining
+a copy of this software and associated documentation files (the
+"Software"), to deal in the Software without restriction, including
+without limitation the rights to use, copy, modify, merge, publish,
+distribute, sublicense, and/or sell copies of the Software, and to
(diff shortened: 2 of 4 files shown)


lastquestion:
I think we should store RP-specific values somewhere in the restaurant profile code. However, if we do have site or design wide values, such as the breakpoints from full desktop to tablet to mobile, those should definitely live here. The name `otkit-layout` seems to make sense, if we do indeed have values that should exist across the design system.

cc @mannionaco cc @lindseyhight

lastquestion:
I'm also not sure if we should store the breakpoint value as a string or number. It is probably best to store it as an number because then we can do (easy) math on it in both JS and CSS (and other formats, too).

ericz82:
For now just have `breakpoint-desktop-tablet` as a token then? I currently have that as a ticket.

lastquestion:
I'd like to have design's input here and ask what from their perspective is going to be shared across pages before we decide on what to introduce.

lindseyhight:
We have designed for breakpoints at tablet and mobile and intend to utilize these across the site in future.

ericz82:
For now I will just have `breakpoint-desktop-tablet`

lastquestion:
This seems reasonable, @nickbalestra what's your thoughts on putting it in a new module? Shall we just put it in `spacing`?

nickbalestra:
Great job @ericz82 ! What about simply calling it `otkit-breakpoints`.
I think it totally make sense to have it.

ericz82:
Sure, since we are only going to be using it for the site-wide breakpoints, `otkit-breakpoints` works for the name.

ericz82:
I currently don't know the tablet to mobile breakpoint value or if we would even have a use for it here, but @lindseyhight or any other designer can add it if it becomes appropriate.

nickbalestra:
@ericz82 I guess you would want to put in your personal details for the `author` field in the manifest :)

lastquestion:
Looks great, let's merge this.

PR #42, opened 2017-09-14
Added Blue Malibu colour for Guest Centre Admin

ronniechong:
This is the primary blue color that is mainly used in Guest Center Admin

### DIFF
--- a/OTTheme/aliases.yml
+++ b/OTTheme/aliases.yml
@@ -26,6 +26,8 @@ aliases:
     value: "#B1D9E7"
   color-juneberry:
     value: "#356273"
+  color-malibu:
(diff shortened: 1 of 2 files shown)


nickbalestra:
Thank you @ronniechong !

lastquestion:
I wonder if we should have put it in OTTheme or OTKit, though? And perhaps we should have design check it over - maybe the name should be like gc-primary? I dunno.

nickbalestra:
I think it was safe to put it in OTTheme. 
Once design will start integrate Restaurant specific decisions, those could definitely go in otkit. @lindseyhight @mannionaco what do you think?

PR #43, opened 2017-09-14
[DX-157] Contributing docs

nickbalestra:


CONTRIBUTING.md
```diff
@@ -0,0 +1,21 @@
...
+2. Clone the repo. If it was already cloned, then git pull to get the latest from master.
+3. Run `npm install` before anything else, and wait.
+4. Write code.
+5. Make sure tests pass (`npm run test`) and make a pull request against the master branch. If you are adding a new token, make sure to update the related READMEs ([main radme](https://github.com/opentable/design-tokens/blob/master/README.md) and design-system specific: [ottheme readme](https://github.com/opentable/design-tokens/blob/master/OTTheme/README.md) or [otkit readme](https://github.com/opentable/design-tokens/blob/master/OTKit/README.md) )
+
+
+# To release new versions
+
+1. To check what needs to be published run `npm run updated`.
+2. Make sure you are listed among the publishers for the related packages you need to release and that you logged in with that listed account to npm.
```
karuto: Can you add a link to a NPM package with "Collaborators" section, for example https://www.npmjs.com/package/otkit-typography. I am concerned that people in OT may be confused about where this "list of the publishers for the related packages" is. We should say explicitly that we are talking about the open-source NPM publisher and NPM accounts here.

nickbalestra:
Thanks for the input @karuto, good point. I've added some more info and a how to check specifically for the public-npm contributors of a specific token.

karuto:
@nickbalestra Thanks a lot, great work!

PR #46, opened 2017-09-20
RUSH-1085 Fix types of font weights, add docs about preview & debug

karuto:
### Overview
This PR changes the type of `font-weight-normal` and `font-weight-bold` to a special "raw" type instead of string, so that values can be parsed by Theo correctly without quotation marks around. This will ensure font weight is properly calculated in start page.

Browser won't be able to parse it if type is "string":
image

This PR also adds some documentation on how to preview and debug a token's generated values.

### Special note on the type `raw`

Theo, the design token parser we use, has a preset of types that each maps to a special transformer to do processing. For example, type `string` adds quotation marks around, type `color` does RGBA format conversion, etc. The new type I defined here, `raw` is actually NOT in the supported Theo transformer list, which means it will preserve the value as is, achieving what we wanted.

### Ticket
https://opentable.atlassian.net/browse/RUSH-1085

### DIFF
--- a/OTKit/README.md
+++ b/OTKit/README.md
@@ -39,6 +39,14 @@ import color from 'otkit-colors/token.common.js';
 @import '../node_modules/ottheme-colors/token.scss';
 ```
 
+Preview & debug the token:
+
+Executing `npm run build` will generate the token values in each token's folder, such as `token.scss` or other available formats you specified.
+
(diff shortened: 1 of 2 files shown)


karuto:
@nickbalestra I fixed the commit and changed to `raw`. Left a paragraph in PR description explaining our thought process as we discussed for future reference.

PR #48, opened 2017-10-04
[DO NOT MERGE][RUSH-886] First batch of desktop typography values

karuto:
### Overview
This PR is based on @lastquestion's PR https://github.com/opentable/design-tokens/pull/32. The discussion related to this ticket can be found [here](https://github.com/opentable/design-tokens/issues/26).

The purpose of this PR is to introduce a batch of platform specific values that are composed of the "original", platform agnostic values.

This PR also changes the previously "raw' types to "size" for contextuality. It does not affect the end result after compiling.

This is a work in progress PR and should not be reviewed yet.

### Ticket
https://opentable.atlassian.net/browse/RUSH-886

### DIFF
--- a/OTKit/otkit-typography/token.yml
+++ b/OTKit/otkit-typography/token.yml
@@ -30,17 +30,81 @@ props:
     type: "size"
     value: "{!line-height-xlarge}"
   font-weight-normal:
-    type: "raw"
+    type: "size"
     value: "{!font-weight-normal}"
   font-weight-medium:
     type: "number"
     value: "{!font-weight-medium}"
   font-weight-bold:
-    type: "raw"
+    type: "size"
     value: "{!font-weight-bold}"
   font-name:
     type: "string"
     value: "{!font-name}"
+
+  heading-xlarge-font-size:
+    type: "size"
+    value: "{!font-size-xlarge}"
+  heading-xlarge-font-weight:
+    type: "size"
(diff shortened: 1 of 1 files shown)


OTKit/otkit-typography/token.yml
```diff
@@ -30,18 +30,82 @@ props:
...
+    value: "{!font-weight-bold}"
+  heading-xlarge-line-height:
+    type: "size"
+    value: "{!line-height-xlarge}"
+  heading-xlarge-color:
+    type: "color"
+    value: "{!color-gray-primary}"
+  heading-large-font-size:
+    type: "size"
+    value: {!font-size-large}"
```
matteofigus: @karuto there is a missing double quote here `"{!font....`

nickbalestra:
I though @karuto explain me that he was to use 'raw' for everything that has not a type being used, to make it easier for designer to understand, this seems to a type that is not actually used instead. What designers think of this?

karuto:
@nickbalestra Umm, I think I actually added the raw -> size change after our discussion.

nickbalestra:
Yes @karuto , as we discussed I think that raw is way more clear then size, let's stick to that :)

karuto:
Thanks @nickbalestra , I will refactor this and then let you review again.

karuto:
Closing in favor of PR https://github.com/opentable/design-tokens/pull/64

PR #50, opened 2017-10-06
Add color-white: #FFFFFF

francesyun:
Add color-white, a new design token. Mostly used for button text for timeslots and Button Primary.


lastquestion:
Uh, @karuto , this didn't actually publish the color anywhere.

PR #54, opened 2017-10-11
[DO NOT MERGE][MINOR] Modified color-primary-disabled. Updated color-primary-active

francesyun:
- Deleted color-primary-disabled in aliases.yml, as it doesn't have it's own independent value (it is equivalent to color-gray-secondary). Developers can still use the variable under its new definition, as defined in token.yml in otkit-colors.

- Updated color-primary-active. The hex value was out of date.

### DIFF
--- a/OTKit/aliases.yml
+++ b/OTKit/aliases.yml
@@ -9,16 +9,14 @@ aliases:
   color-primary:
     value: "#DA3743"
   color-primary-active:
-    value: "#FB665F"
-  color-primary-disabled:
-    value: "rgba(218, 55, 67, 0.40)"
+    value: "#AE2C35"
 
   # Primary Gray, used on all text sizes
   # =============================================
   color-gray-primary:
     value: "#333333"
 
-  # Secondary Gray, used on smaller font sizes
+  # Secondary Gray, used on smaller font sizes and for disabled primary buttons
   # =============================================
   color-gray-secondary:
     value: "#717171"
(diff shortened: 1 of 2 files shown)


karuto:
Can we talk about this. I want to bring up a discussion about functional naming vs color naming.

We used to do this thing in the old opentable-theme where we have a set of "pure color variables" and we remap them to all these "functional variables" such as this:

image

Where the left hand side is a functional variable that has no color implication (e.g. you can remap to any other color and it still makes sense) and the right hand side is pure color without any functional implication. All is good.

The benefit of functional variables from a dev perspective is that it's more intuitive if the "set" of colors is complete and variable names are predictable. For example if I'm writing a button. I would expect `color-button-active`, `color-button-disabled`, `color-button-hover` and so on. This forms a whole set. The naming convention is aligned with the states of the UI.

So in that sense I would actually advocate to keep "color-primary-disabled" and just have it map to the gray, like what we are doing in opentable-theme, instead of removing it completely.

Maybe consider "completing the set of colors to match the possible states of the UI" for other applicable color variables, and leave the "pure color variables" outside of the functional world? For example, if you add "color-primary-disabled" back, you can move the comment about "color-gray-secondary" up as well.

Wondering if this makes sense, curious to hear your opinions! @francesyun

karuto:
Let's put this PR on hold until we reach an agreement on color namings.

karuto:
Closing this in favor of https://github.com/opentable/design-tokens/pull/75

PR #56, opened 2017-10-24
[MW-1975] Use Pixels for OTKit. (And a few README clean-ups)

cbeley:


cbeley:
Few open questions:

The default behavior for pxtorem is to only apply pixel to rem conversion for ['font', 'font-size', 'line-height', 'letter-spacing'].  Should we set it for other things?  

Potentially less relevant if we're just going to transform font sizes, but should we set `minPixelValue` and to what.  `minPixelValue` determines whether or not to do the conversion.  For example, if we set it to `2`, anything under 2px will stay as px.

Are there any other defaults listed at https://github.com/cuth/postcss-pxtorem#options that we should override?

Finally, is having a recommended configuration in the README enough, or should we export our recommended configuration somehow (I think having it in the README is good enough for now, but as we share some of our build configuration, there is more we could do here).

OTKit/README.md
```diff
@@ -19,15 +19,41 @@
...
+```
+
+Then add it to your [postcss](https://github.com/postcss/postcss) processors with the following recommend configuration:
+
+```
+import pxtorem from 'postcss-pxtorem';
+
+const processors = [
+  pxtorem({
+    rootValue: 16,
```
cbeley: While this is the default value for pxToRem, this is just a critical thing, that I didn't want to risk pxToRem changing the default one day for whatever reason.

OTKit/README.md
```diff
@@ -19,15 +19,41 @@
...
+    selectorBlackList: [/^html$/],
+    replace: true
```
cbeley: This just omits pixel fallbacks for browsers that don't support rems (pre IE11).  Based on my understanding of what browsers we support at OpenTable, this should be a sane suggestion.

cbeley:
Also, I did not change the values in OTTheme, since I wasn't sure what the story is there.  Should we change that now, or potentially later?  I'm not very familiar with OTTheme.

cbeley:
Other open question:  Should we export our rootValue as a design token?  I don't see it anywhere currently.

nickbalestra:
As there are some Qs here and there, I'll try to group them here:

- `OTTheme`: I wouldn't do it atm, as we'll have to adjust the build process on every build-site, being that our `legacy` design I think that there is no nreal eed/benefit to do that.
- `rootValue`: I think it would be good to include that with it, as the two value are strictly related/dependent, perhaps with a less tool-dependent name (ie" `font-size-default` or `font-size-base` ,...)?
- Should we set it for other things? Yes, I personally feel that consistency will be good, so that we either rely on px here or on rem, what you folks thinks @lastquestion @karuto @matteofigus. Would love to hear what designer suggest @mannionaco @lindseyhight @francesyun ?

OTKit/aliases.yml
```diff
@@ -37,40 +37,40 @@ aliases:
...
   font-size-large:
-    value: "1.5rem"
+    value: "24px"
   font-size-xlarge:
-    value: "3rem"
+    value: "48px"
   # line heights seem arbitrary because they are mapped to specific font sizes
   line-height-xsmall:
-    value: "1.43"
+    value: "16px"
```
karuto: You are changing the behavior of line heights with these changes? These line height values used to be numerical which means they scale (multiply) with the corresponding font size to get the final line height in px value.

If you do the line height calculation based on your own pixel values changed above, none of them seem to line up. For example, you set `font-size-xsmall: 14px`, then its line height should be `14px * 1.43 = 20.02px`

cbeley: Ah, that is totally my bad.  I did not realize that and replaced them with what is in the design document.

I don't think this is super clear though and am still a little confused.  This doesn't seem to line up with what design has speced out for the line-heights.  How were these ratios determined to begin with?

cbeley:
@nickbalestra I like `font-size-base` the best, but I think both are good options.  I also agree that calling it root-value isn't the best generic term.

Anyone have a particular preference for naming?

OTKit/aliases.yml
cbeley: Here's the current design spec:

image

cbeley:
@nickbalestra As for setting it elsewhere, it's not quite a consistency thing.  The problem with setting rems for non-fonts is that it may render things in unexpected ways (That's why there are options to say, don't convert things under X px to rems).  For padding, rems may be ok, but for borders, it may make sense to universally keep those as pixels so the design doesn't ever look shrunk in a strange way.

I think there are others that can talk about this issue more though (since I've seen some others post comments about this).

Either way, the beauty of using a tool for this, is we can automate this process of using pixels and rems where appropriate.

cbeley:
Also, just to be super clear (since re-reading your comment makes me think there may be a miss-understanding) is that we'll still use pixels everywhere in this repo.  This is a question of how to configure the pxToRem tool to make the right decision of whether to use pixels or rems in various situations.

cbeley:
One example:  we can't convert 1px borders to rems, since there could be cases where the border is no longer visible.

OTKit/aliases.yml
cbeley: Regarding the line-heights, talked with Eric a bit, and I think it should be safe for us to simply update this to be in line with what design has speced out for line heights in pixels.  Everyone is still just doing things like `line-height: $lineHeightSmall`.

I'm also not sure how the current ratios were set up, since they don't seem to match up with the spec either way after doing the math.

lastquestion: I think it's probably ok as this will be a major version bump anyway, so we can adjust the line heights to match the current spec and verify the pages as we bump up to the major version.

morewry:
Ratio `line-height` does have its advantages if it fits the typographic rhythm to design's satisfaction. I think that for initial implementations, however, it does make sense to keep it simple.

As a note for any later exploration of that: I don't know that pxtorem supports it, but the math to converting px to ratio, rem, and em is the same; the only difference is that em and ratio aren't absolutely determined by _root_ font size, so if you're setting or inheriting a different `font-size` you have to use that specific size to divide instead of the root one.

cbeley:
So, considering the silence, I'll put out updates to this PR as follows:

 * Keep the pixel line heights I set (and clean -up the confusing now irrelevant comment)
 * Add `font-size-base` (seems like the clearest suggestion and no one else has chimed in with other feelings).

There are open questions about pixel to rems settings, but I think the settings + defaults I have now are good enough to get started.  I can open another issue to discuss more at length if we think we should stray from the defaults (which would have some implications and other considerations...same as if we were doing this manually though).

OTKit/README.md
lastquestion: can't we get this value directly from the design token itself?

OTKit/README.md
```diff
@@ -19,15 +19,41 @@
...
+```
+
+Note that depending on how you use postcss, your processors array may be in a JSON configuration file instead.
+
+If you're not already using [postcss](https://github.com/postcss/postcss), you will either have to use it as an additional build step after you transform your CSS with your current tool or use it as a complete replacement for your current CSS transformation process.
+
 ### Use the token
 
-A Token exposes multiple available formats (listed above). **The format need to be excplicitely referenced upon requiring/importing the token:**
+A Token exposes multiple available formats (listed above). **The format needs to be explicitly referenced upon requiring/importing the token:**
```
lastquestion: 👍

OTKit/README.md
cbeley: Yup, though with one caveat.  People will not be able to do this if they use the JSON configuration approach for postcss.  However, i think that will be relatively uncommon for most projects at OpenTable.

lastquestion: Feel free to do this in this PR or in a follow up

lastquestion:
Sorry, please do export the `font-size-base` and then this PR is approved (i hit approve too fast).

cbeley:
@lastquestion Everything is updated.  Also added a bit more documentation on base sizes.

OTKit/aliases.yml
karuto: You should talk to @francesyun because the spec is outdated and the px line height aint right @cbeley

karuto:
Please merge only after resolving the line height issue with @francesyun and Kellen - thanks in advance 🤓

cbeley:
@karuto I'll double check, but I don't think there is any issue to resolve.  It should be pixels last I chatted with him.

OTKit/aliases.yml
```diff
@@ -37,40 +37,42 @@ aliases:
...
```
mannionaco: xsmall - line height = 20px (not 16) 

This is the only change.

cbeley:
@lastquestion @mannionaco All look good to you?  I think we're ready to merge.

PR #70, opened 2017-11-03
Use the correct registry for publishing

matteofigus:
This should just make things easier

### DIFF
--- a/OTKit/README.md
+++ b/OTKit/README.md
@@ -52,8 +52,6 @@ If you are using a token in your project, you can execute `npm link '<token-name
 
 ### Publish
 
-A token needs to be published in order for the changes to take effect. To publish a token, navigate to the token's directory, for example `./OTKit/otkit-colors/`, then run `npm publish otkit-colors`.
-
 You need to have an NPM account and be among the "Collaborators" list on the official NPM package page, for example  `https://www.npmjs.com/package/otkit-colors`, to publish. Contact one of the collaborators if you would like to be added.
 
 
--- a/OTKit/README.md
+++ b/OTKit/README.md
@@ -56,9 +56,6 @@ A token needs to be published in order for the changes to take effect. To publis
 
 You need to have an NPM account and be among the "Collaborators" list on the official NPM package page, for example  `https://www.npmjs.com/package/otkit-colors`, to publish. Contact one of the collaborators if you would like to be added.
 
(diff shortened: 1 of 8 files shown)


karuto:
@matteofigus This is great, is it possible to set this in the package.json in the root instead of in each token? Hmm, maybe that won't work because we execute npm publish inside a token's folder?

matteofigus:
I don't know, I think this is what is recommended. Another approach could be to just have a .npmrc in the root using the public? But I think we should try to see if it really works.

Perhaps we can do a pair on Monday for a live spike.

nickbalestra:
@matteofigus as you at it, could you change this? https://github.com/opentable/design-tokens/pull/70/files#diff-23097fc8f5cd2cc4db32a850309a2adfR55 I feel we should rely on lerna for publishing instead of doing manual npm publish from within each package, as that defeat the very first reason on having lerna. What you think?

matteofigus:
@nickbalestra done.

karuto:
@nickbalestra We should add instructions in README about how to use lerna's publish. Right now, there is no information about that, that's why everyone's still assuming they should default to npm

nickbalestra:
Not sure that info should be int he otkit specific readme as the publishing process is general to the design-tokens system and repo. In fact we already mention it in the [contributing doc ](https://github.com/opentable/design-tokens/blob/master/CONTRIBUTING.md)

karuto:
@nickbalestra Thanks for the link, but as far as I can tell, the [CONTRIBUTING.MD](https://github.com/opentable/design-tokens/blob/master/CONTRIBUTING.md#to-release-new-versions) still asks people to use `npm run publish` instead of lerna? We should update it here then.

karuto:
@matteofigus We should merge this after the conflict is fixed. I think it's a good step forward than what we currently have. Let me know if you are busy and I can help fixing the conflict for you.

nickbalestra:
@karuto -> indeed and thats why: https://github.com/opentable/design-tokens/blob/master/package.json#L13. 

Although the fact that you didn't know, tells me that perhaps we should call it different then `publish` any thoughts for what could be more intuitive?

karuto:
@nickbalestra Yes I just realized it is `npm run publish` rather than `npm publish` after you mentioned it. The simple thing could be just add a line in that section calling out the fact that `run publish` is powered by Lerna behind the scenes.

Issue #68, opened 2017-11-01
Excessive Aliasing & What We Should Do About It


cbeley:
So, right now, we are aliasing *everything* in `aliases.yml`.  This is a bit confusing.

I propose we only move things into `aliases.yml` if and only if the value needs to be shared.  Otherwise, it should be inlined in the `token.yml` file, so it's very clear what the values are for the given token file.

If, down the road, it needs to be shared, then you can move it into `aliases.yml`.

Vote with the two posts below this, and let's chat about why this may or may not be a good/bad idea. :)

cbeley:
Thumbs up this if you think we should only add things to alias.yml if and only if a value is used in more than one place.

cbeley:
**Thumbs up** this if you think we should continue the current pattern of putting all raw values in `alias.yml`.

karuto:
I support the idea of only putting values that need to be shared among multiple tokens into aliases.

Please vote if you care about this issue. If a comment has more than 3 votes, we will conclude this issue 7 days after its opening, on Nov 8th 2017 in US time. This is based on our moderation proposal in #66.

cbeley:
@karuto Assuming we don't take the functional color names approach, I think the point may be a bit moot.  I would expect shared values to be on the rarer side.

cbeley:
I'd say we can just clean it up in one big ticket.  Should be fairly quick to do.

I could probably find some time for it in my next sprint even.   Better to do it now than later before things get too crazy.

matteofigus:
Before doing this, I would like some input from design. All the thumbs up are from engineers - just want to be sure @mannionaco @francesyun @lindseyhight have seen this thread?

matteofigus:
@cbeley we just had a quick meeting with @francesyun and other designers, and discovered the engineering naming wasn't super inclusive for designers. Can we include a little example of a change, so they can contribute to the conversation? 👍

cbeley:
I got this scheduled for my coming sprint: https://opentable.atlassian.net/browse/MW-2038 .

As for example, see https://github.com/opentable/design-tokens/blob/master/OTKit/otkit-typography/token.yml#L31 .  See how, instead of the value being there, it's actually in https://github.com/opentable/design-tokens/blob/master/OTKit/aliases.yml#L78 .  With the proposed change, we'll inline it, so in token.yml, it looks like:

```
line-height-xlarge:
    type: "size"
    value: "48px"
```

[...]

PR #73, opened 2017-11-15
[MW-2038] Remove unnecessary aliasing.

cbeley:
Previously discussed in https://github.com/opentable/design-tokens/issues/68.

### DIFF
--- a/OTKit/otkit-colors/token.yml
+++ b/OTKit/otkit-colors/token.yml
@@ -1,21 +1,38 @@
 props:
+  # Primary Action
+  # This color is only used on objects, backgrounds with large text (24px or larger, bold), and will be used as a color to guide users towards actionable content.
+  # =============================================
   color-primary:
-    value: "{!color-primary}"
+    value: "#DA3743"
   color-primary-active:
-    value: "{!color-primary-active}"
+    value: "#FB665F"
   color-primary-disabled:
-    value: "{!color-primary-disabled}"
-  color-gray-primary:
-    value: "{!color-gray-primary}"
-  color-gray-secondary:
-    value: "{!color-gray-secondary}"
-  color-gray-utility:
-    value: "{!color-gray-utility}"
-  color-white:
-    value: "{!color-white}"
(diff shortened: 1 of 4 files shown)


cbeley:
I didn't do it in this PR, but if the plan going forward is to share values between otkit-desktop-typography and otkit-typography (which, I'm a little confused about right now and why we are doing the aliasing to begin with with all the different names...need to catch up with design), it may make more sense to create a file 'typography-aliases.yml' instead and import from that to be clearer.

cbeley:
Also, if we're going the functional route for colors, we could do a similar thing: color-aliases.yml to keep everything organized.

cbeley:
So, I'll be out after Friday until the 29th.  Anything we want to change with this PR in particular?  It's moves a lot of things around, so would be better to get this out sooner than later.

Things like creating multiple aliases files can occur in other PR's.

OTKit/otkit-colors/token.yml
```diff
@@ -1,21 +1,38 @@
...
+  # =============================================
   color-gray-utility:
-    value: "{!color-gray-utility}"
-  color-white:
-    value: "{!color-white}"
+    value: "#E1E1E1"
+
+  # Lightest gray, used for backgrounds for the page and backgrounds for active/hover states
+  # =============================================
+  color-gray-active:
```
karuto: Thanks for adding this. This is another problem with excessive aliasing that you discovered: humans are not perfect and we sometimes forget to map a variable, and the system doesn't remind us for that. 👍

karuto:
@cbeley I think we can go ahead and merge today once I hear your feedback, but this should be a major version bump. What do you think? Would you be able to publish?

matteofigus:
@karuto I can publish when merged, just let me know.

cbeley:
I think we are still good to go?  Is there something new that may have happened while I was out I'm not aware of?

karuto:
@cbeley Not much, I think everyone's on the same page. I'm merging.

PR #88, opened 2018-01-04
[RUSH-1363] Borders token + housekeeping

karuto:
### Overview
This PR introduces the borders token, which will store border related attributes, such as border radius, border width, border color, border style, etc. 

This PR just creates the token to lay the ground, we will work with designers to fill more values in there.

This PR also does some housekeeping:
 - removes the last remaining personal developer info from author field;
 - removes the homepage field in `package.json` because individual token doesn't have readme anymore;
 - adjusts the order of tokens in main readme to be consistent with the alphabetical order of folders.

### DIFF
--- a/OTKit/otkit-borders/package.json
+++ b/OTKit/otkit-borders/package.json
@@ -1,7 +1,7 @@
 {
-  "name": "otkit-breakpoints",
-  "version": "2.0.1",
-  "description": "OpenTable breakpoints design token",
+  "name": "otkit-borders",
+  "version": "0.0.1",
+  "description": "OpenTable borders design token",
   "author": {
     "email": "design-tokens@opentable.com"
   },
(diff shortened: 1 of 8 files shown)


karuto:
@matteofigus what should I do for publishing when I create a new token? Should I run normal `npm publish` or use `lerna`(`npm run publish`)? I don't think the docs have this scenario explained, I am happy to add it in once you teach me. 😃

OTKit/otkit-borders/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "otkit-borders",
+  "version": "1.0.0",
```
matteofigus: The convention I found good, is to make a new package having `0.0.1` - once merged, we can publish as `1.0.0` with a `npm run publish` on the root - if you have 1.0.0 it forces you to bump it to something like 1.0.1 which is annoying

karuto: Ok that makes sense. I will put this note in readme for people want to create new tokens in future.

karuto:
@matteofigus can you review again, I fixed the version number.

Issue #91, opened 2018-01-18
Additional tokens for ticketing


francesyun:
On the [ticketed experience modal](https://opentable.invisionapp.com/share/MBFAVB5ES#/273305247_RPP4-Longstring), the font size decreases if the ticket name exceeds 23 characters. This smaller font size (font-size: 32px, line-height: 40px) isn't a token yet. It would be between our current font-size-xlarge (font-size: 48px) and font-size-large (font-size: 24px). 

➡️ What are your thoughts on how we should approach the naming? 
**Suggestion**: Change font-size-xlarge to font-size-xxlarge, and add 32px as font-size-xlarge.

Also, we would like to add letter-spacing: 1px to the design tokens, since we plan to use it in several places when we do ALL CAPS. There are no other expected changes to letter-spacing at this time. 

➡️ Do you have thoughts about what to name this token? 
**Suggestion**: letter-spacing-stretch

@karuto @mannionaco

mannionaco:
@francesyun @karuto I don't have a strong opinion here, if we are going forward with this font size it seems that it fills the "gap" and no other introductions will be needed. 

Therefore, @francesyun, your first suggestion, pending @karuto's feedback, I can get behind. Where we bump xl to xxl and change xl to 32px....so we are editing font-size-xlarge to 32px and adding a new token for xxl at 48px. Sound good?

karuto:
@francesyun @mannionaco One problem I can see with Frances' suggestion is that we need to replace every current usage of font-size-xlarge because the contract has changed. This is doable, but a bit tedious as many projects use font-size-xlarge now, not just rest profile.

I support the idea of adding letter spacing as a token, it is just a matter of time. I'm thinking whether we should bundle letter spacings inside the existing typography token or create a new one. I guess it depends on how explicit we want these "atom level" tokens to be. Right now, the typography token stores only 3 values: line-height, font-size and font-weight.

I don't have opinion on the naming of letter spacing, as long as it makes sense inside your broad design system! 🤗

karuto:
Digging this issue up since @francesyun is doing another round of design review for the ticketing experience modal, which you can see over here: http://oc-registry-pp-sf.otenv.com/v2/ticketing-modal-web/1.0.6/~preview?tid=1001

[...]

PR #97, opened 2018-02-07
Create mobile web typography token, change xlarge + xxlarge

karuto:
This PR is a result of discussion thread https://github.com/opentable/design-tokens/issues/91, seems like we reach agreement on the following action items:

* create a new xxlarge values in https://github.com/opentable/design-tokens/blob/master/OTKit/otkit-typography/token.yml
* change the xlarge values in https://github.com/opentable/design-tokens/blob/master/OTKit/otkit-typography/token.yml
* create a new xxlarge font group in https://github.com/opentable/design-tokens/blob/master/OTKit/otkit-desktop-typography/token.yml
* create a new set of font groups (that are not unique, e.g. are same as web) in a new package otkit-mobile-typography

The mobile web font group values are from the typography stack image: https://github.com/opentable/design-tokens/issues/91#issuecomment-363507564.

Questions for @mannionaco and @francesyun:
 * We gave semantic names to font groups in desktop web, such as `heading-[size]`, `bodytext-[size]` etc. Can you advise on the mobile web ones?
 * In the typography stack image, seems like mobile web and desktop web have different definitions of xsmall and small. I would like to hear your inputs before putting them in. Inside ` OTKit/otkit-mobileweb-typography`, we have 2 ways of assigning them:

1. We reuse the shared pool of already defined variables, but may cause confusion:
```
  bodytext-small-font-size:
    type: "size"
    value: "{!font-size-xsmall}"
```
2. We define a new value locally, straight forward but missing the reusability:
```
  bodytext-small-font-size:
    type: "size"
    value: "14px"
```
I personally leaning towards the approach in 2, but this makes me question the purpose of `OTKit/aliases.yml` because it exists in a desktop-centric worldview.

Speaking of which, we should think about what's the purpose of `otkit-typography` token going forward. Fear that it will turn into legacy, as we create more of these platform specific tokens. Feel free to chime in from a design perspective @francesyun and @mannionaco. I'll create a separate issue if people would like to discuss.


karuto:
This PR will not be worked on until we get a clear feedback from design about the future of platform typography, the balance between unification vs divergence, etc.

mannionaco:
We have chosen to have two type ramps. One for Web (desktop, tablet) and a second for Mobile (mWeb, iOS, Android). Web is composed of only our brand font, Brandon Text. Mobile is composed of our brand font + system font. 

We decided to use system fonts within mobile because we know that audience quite well (mostly iOS users) and knowing that system fonts are built specifically for those mobile devices made it easier to decide on that direction, let alone potential performance and experiential (ex. "the font flash") benefits. On a level of value and impact Brandon Text's style is noticeable at larger sizes, less noticeable at smaller sizes.

I see a future where we may merge this POV with our Web typography, but as I see it today if we were to use system fonts within Web the variability is exponentially more difficult to control compared to mobile (ex. users seeing Arial, a font I can't subjectively allow to happen). If I am wrong let's work on a merged font library at a later date with a more robust font stack (ex. LinkedIn). 

I don't see any additional type ramps being created within the consumer experience, nor do I see the restaurant product teams using a different set. Though we will evolve this, but creating isolated font libraries to just create them goes against our unification goal. However, we will deviate with intention with appropriate rational.

When it comes to the different sizes and names demonstrated here:
https://opentable.invisionapp.com/boards/E53ARCZ736G/

I wasn't sure what would be helpful and used this to demonstrate what is shared and how System Fonts (ex. San Francisco, Android) are optically larger (roughly 15%) than our brand font Brandon Text. I don't think they need to be different. I wasn't sure what to do tbh.

I moved around the comparison graphic: 
https://opentable.invisionapp.com/boards/BR3ARCZ8VJG/

Is this helpful? Small in Mobile serves the same purpose as Small in Desktop. 

Does this information help inform the decision you mention above? May there is a mobile font library and desktop font library?

char1e5:
cc @lastquestion @matteofigus 😃

lastquestion:
@char1e5 please see the issue #142 :)

karuto:
Love the conversations we are having, but I realized that I'm trying to do too much in this one PR, which caused it to turn stale. I am closing this in favor of https://github.com/opentable/design-tokens/pull/145, which does half of what this PR wants to do, but will get merged faster.

PR #99, opened 2018-02-13
Style guide: Initial typography styling

francesyun:
- Add lodash dependency
- Add global font face (Brandon Text)
- Display typography groupings 

screen shot 2018-02-12 at 4 34 43 pm

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ node_modules
 *.common.js
 .cache
 public
+package-lock.json
--- a/.gitignore
+++ b/.gitignore
@@ -6,4 +6,3 @@ node_modules
 *.common.js
 .cache
 public
-package-lock.json
--- a/style-guide/src/pages/otkit-desktop-typography.js
+++ b/style-guide/src/pages/otkit-desktop-typography.js
@@ -1,41 +1,41 @@
-import React from "react";
-import Link from "gatsby-link";
-import _ from "lodash";
+import React from 'react';
+import Link from 'gatsby-link';
+import _ from 'lodash';
 
-import token from "otkit-desktop-typography/token.common";
-import styles from "../styles/otkit-desktop-typography.module.css";
+import token from 'otkit-desktop-typography/token.common';
+import styles from '../styles/otkit-desktop-typography.module.css';
 
 const typographyGroups = [
   {
-    name: "heading-xlarge",
+    name: 'heading-xlarge',
     isAlternate: false
   },
   {
-    name: "heading-large",
+    name: 'heading-large',
     isAlternate: false
   },
   {
-    name: "heading-large",
+    name: 'heading-large',
     isAlternate: true
   },
   {
-    name: "heading-medium",
+    name: 'heading-medium',
     isAlternate: false
   },
   {
-    name: "subheading-small",
+    name: 'subheading-small',
     isAlternate: false
   },
   {
-    name: "bodytext-small",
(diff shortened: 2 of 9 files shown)


lastquestion:
Screenshot ;)

.gitignore
```diff
@@ -5,4 +5,5 @@ node_modules
 *.cssmodules.css
 *.common.js
 .cache
-public
\ No newline at end of file
+public
+package-lock.json
```
lastquestion: no, this should be checked in.

lastquestion: no, this should be checked in.

style-guide/src/pages/otkit-desktop-typography.js
```diff
@@ -1,14 +1,65 @@
 import React from 'react';
 import Link from 'gatsby-link';
+import _ from 'lodash';
 
 import token from 'otkit-desktop-typography/token.common';
-import exampleStyle from '../styles/example.module.css';
+import styles from '../styles/otkit-desktop-typography.module.css';
+
+const typographyGroups = [{
```
lastquestion: fix the spacing here

style-guide/src/styles/otkit-desktop-typography.module.css
```diff
@@ -0,0 +1,10 @@
+@value spacing-large from 'otkit-spacing/token.cssmodules.css';
+
+.font-column {
+	width: 500px;
```
lastquestion: fixing the spacing here

style-guide/src/pages/otkit-desktop-typography.js
```diff
@@ -1,66 +1,71 @@
-import React from 'react';
-import Link from 'gatsby-link';
-import _ from 'lodash';
+import React from "react";
```
lastquestion: these should be single quotes everywhere, please. Also, I know this isn't your fault, because we don't have a linter yet, and these kinds of rules should be encoded by the linter. Sorry >_<

style-guide/src/styles/otkit-desktop-typography.module.css
lastquestion: Actually, I'm not sure, if we are going to use 8 spaces or 4 or 2, but I think generally we are using 2, so I'd like to claim we should use that. ^_^

lastquestion:
@matteofigus can we claim node 8 / npm 5 for this repo and add it to the readme / package.json engine field?

style-guide/src/pages/otkit-desktop-typography.js
matteofigus: We used prettier with default settings...sorry let me change the setting. We can definitely automate later on with pre-commit prettier...

lastquestion: yeah, that sounds good 👍

matteofigus:
@lastquestion we already use that https://github.com/opentable/design-tokens/blob/master/.travis.yml but I wouldn't put it in the docs being an internal detail...the libraries work with 6 too

lastquestion:
yeah, but the generated `package-lock.json` tells me everyone else was using npm 3/4? ❓  ;)

lastquestion:
Also, I think we should have prettier rules running for the css in the style guide too. :)

matteofigus:
Let me add an issue for that. Nick is gonna integrate it in no time, he loves prettier 💃

PR #103, opened 2018-02-14
Add layout to style guide

francesyun:
- Add basic 2 column layout for navigation (using same column widths as new restaurant profile)
- Add section header component to be shared across pages (header for "Typography", "Colors", etc.)
- Add starter file for colors page

screen shot 2018-02-13 at 5 33 26 pm


style-guide/src/pages/index.js
```diff
@@ -1,17 +1,17 @@
 import React from 'react';
 import Link from 'gatsby-link';
```
ericz82: You can remove this import statement now.

style-guide/src/pages/otkit-colors.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import Link from 'gatsby-link';
```
ericz82: You can remove this import statement.

style-guide/src/pages/otkit-desktop-typography.js
```diff
@@ -1,6 +1,7 @@
 import React from 'react';
 import Link from 'gatsby-link';
```
ericz82: You can remove this import statement now.

style-guide/src/components/section-header/index.js
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+import styles from './styles.module.css';
+
+import token from 'otkit-desktop-typography/token.common';
+
+export default function SectionHeader({ text }) {
+	return <h2 className={styles['section-header']}>{text}</h2>;
```
ericz82: It's unfortunate that CSS Modules in Gatsby does not enable exporting class names in camel case.

ericz82:
https://github.com/opentable/design-tokens/pull/103/files#diff-20114c05cb418938ef0f27645734aaecR61

BTW, you should be adding keys to elements when mapping over an array.

style-guide/src/components/section-header/index.js
nickbalestra: I think that you should be able to extend the configuration of the gatsby webpack to add support for that if you feel the need - https://github.com/gatsbyjs/gatsby/blob/master/docs/docs/add-custom-webpack-config.md

nickbalestra:
@ericz82 isn't that for React to check for changes when recursing on children? 
As we are not adding nor removing any of the items in the generated list no recursing on those children should ever occur and therefor there should be no need for keys there, also if no key is provided, React add the index by default as the key (again in our case, for the reasons above, I would argue it being an [anti-pattern](https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318), but simply relying on a default as we'll never need React to iterate and check there for now). 

TLDR; So i think we should be safe to omit keys, and simply rely on React defaulting to indexes for them

- [children recursion](https://reactjs.org/docs/reconciliation.html#recursing-on-children)
- [key default to indexes](https://reactjs.org/docs/lists-and-keys.html#keys_

ericz82:
@nickbalestra That makes sense since this is a static website! React may throw warnings but if that's not an issue, then we're good.

style-guide/src/components/section-header/index.js
lastquestion: @nickbalestra , is this PR not rebased against your prettier PR we just merged? Doesn't prettier do 2 spaces by default?

style-guide/src/components/section-header/index.js
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+import styles from './styles.module.css';
+
+import token from 'otkit-desktop-typography/token.common';
```
lastquestion: what's this import for?

style-guide/src/components/section-header/styles.module.css
```diff
@@ -0,0 +1,15 @@
+@value spacing-large, spacing-medium from 'otkit-spacing/token.cssmodules.css';
+@value heading-large-font-size, heading-large-font-weight, heading-large-line-height from 'otkit-desktop-typography/token.cssmodules.css';
+@value color-gray-primary, color-gray-utility from 'otkit-colors/token.cssmodules.css';
+
+.section-header {
+	font-size: heading-large-font-size;
```
lastquestion: can we go with hard 2 spaces for tab ident? 💓

style-guide/src/layouts/index.js
```diff
@@ -0,0 +1,25 @@
...
+export default ({ children }) => (
+	<div className={styles['two-column']}>
+		<div className={styles['nav']}>
+			<Link className={styles['link-secondary']} to="/">
+				<h1>OTKit Style Guide</h1>
+			</Link>
+			<NavLink to="/otkit-desktop-typography/">Typography</NavLink>
+			<NavLink to="/otkit-colors/">Colors</NavLink>
+		</div>
+		<div className={styles['main']}>{children()}</div>
```
lastquestion: I think it's more idiomatic to write `{children}`. You don't execute it. Actually, what does executing it even do? I know that `children` is not really an array, but is it a function...?

style-guide/src/pages/index.js
```diff
@@ -1,17 +1,17 @@
...
 
 export default () => {
   return (
     <div className={styles.body}>
-      <h1>Hello world</h1>
       <p>
-        <Link className={styles.link} to="/otkit-desktop-typography/">
-          link to otkit-desktop-typography
-        </Link>
+        Welcome to the OTKit Style Guide, the place to see our{' '}
```
lastquestion: Why did you need to add a escaped space via `{}` here? was there something wrong with the normal breaking?

style-guide/src/pages/otkit-colors.js
```diff
@@ -0,0 +1,13 @@
...
+import _ from 'lodash';
```
lastquestion: lodash not used

style-guide/src/styles/index.module.css
```diff
@@ -1,8 +1,51 @@
...
+@value color-primary, color-gray-primary from 'otkit-colors/token.cssmodules.css';
+@value font-weight-medium from 'otkit-typography/token.cssmodules.css';
+
+/* local variables */
+@value width-two-column: 992px;
+@value width-nav: 320px;
+@value width-main: 640px;
+
+body {
+	font-family: 'BrandonText', -apple-system, BlinkMacSystemFont, 'Segoe UI',
```
lastquestion: we probably actually 100% wanna see the actual font, not the fallbacks, so I don't know if the full stack here is useful or bad. Actually, we probably want to have a way of seeing the fallbacks live, so we know they look like when the font doesn't load. Another feature for later.

style-guide/src/layouts/index.js
ericz82: @lastquestion This is a feature of Gatsby. See https://www.gatsbyjs.org/tutorial/part-three/#our-first-layout-component

lastquestion: Wow. Thanks, TIL. I __did not know this__ for sure. Interesting design choice...

lastquestion:
@nickbalestra , I realized why there was a `package-lock.json` file in the `style-guide` directory, it's because either I or @francesyun accidentally ran `npm install` manually within the `style-guide` dir. Anyway, I fixed the `package-lock.json` file to use public NPM registry and correct sha, and let `lerna` install the deps inside `style-guide`. 👍

lastquestion:
I'm merging this and manually running the `postpublish` step so we can get the updated styleguide before we publish an actual update to the tokens.

Issue #93, opened 2018-01-18
Adding SVG icons


francesyun:
We would like to add our OTKit SVG icons as design tokens, since the icons are now being used across several pages. Because we don't have a shared repo with the icons, dev teams are simply duplicating the assets. Could you help us get started on this?

@matteofigus @nickbalestra @karuto 

Thanks!

matteofigus:
Hi @francesyun, I am actually not sure they should be a design token, as I've seen this handled commonly at the component's level - so maybe this is the right place instead? https://github.com/opentable/ot-react-ui-components

But let's wait for @karuto's opinion first - I suspect he put some thoughts on that already 👍

mannionaco:
@aselbie sent this link and for what it's worth maybe it will drive some conversation around this topic and how design executes icons:
https://medium.com/@david.gilbertson/icons-as-react-components-de3e33cb8792
https://icomoon.io/app/#/select

karuto:
We have been circulating ideas around this during our past few design system meetings. I've done some research. I think it's safe to assume that within the context of design tokens, exporting a collection of raw SVG paths directly may be the easiest and most straight forward solution.

We can then let the consumer of this token to decide how to use the paths to make some composited solution. For example if a react-based team imports this token, then most likely they'll create a thin component wrapper around it. If a team is using pug then they can embed the paths in a template tag or inline CSS. That's up to them. I think we'll have maximum flexibility if we keep our assumptions minimum here at the token level, so just export paths as-is.

Let me know what you guys think. I can hack together a proof of concept PR to showcase the usage.

nickbalestra:
I agree with the general direction, that will be terrific @karuto , thanks!

lastquestion:
I think exporting the raw SVG path is good. We probably will need another abstraction layer above design tokens to share react compnents / direct upload to CDN / etc.

What I would like to have is:

1. Designer has a SVG file on disk
2. Designer runs a script that transpiles the SVG (and minifies, checks the path names for collisions, etc. - aka something like https://github.com/svg/svgo - and generates a file for that SVG somewhere in directory structure
3. As part of regular `npm run build`, adjust / use theo as normal to generate the tokens.

lastquestion:
I think we can close this and open new issues for the next steps for the icon story. Most of the original stab outlined above was done in #135 (which superceded #106 above)

PR #106, opened 2018-02-14
[RUSH-1675][WIP] Add OTKit-icons token and icon generation pipeline

karuto:
### Overview
This PR is a result of the SVG icons [discussion](https://github.com/opentable/design-tokens/issues/93) - let's introduce a designer-friendly solution to standardize our SVG icon usages.

Here's my proposed process:
 1. Drop any normal SVG files, like `ic_information.svg`, into the `assets/` folder.
 2. Type `npm run build-icons` to trigger the script to build the `OTKit/otkit-icons/token.yml` automatically. **The file name will become the key, and the optimized / standardized SVG content will be the value.**
3. Publish the token as normal. Our standard `theo` pipeline will publish it as a `common.js` format file.
4. Developer can use this token in a JavaScript project like so:

```javascript
import React from 'react';
import icons from 'otkit-icons/token.common.js';

export default class RandomComponent extends React.PureComponent {

  render() {
    return (
      <div>
      {icons.ic_information} // here's the icon
      </div>
    );
  };

}
```

**NOTE: Still work in progress. Features I plan to add:**
- [x] Pick a node yaml generator to automate the yaml generation
- [ ] Adjust svgo so that it removes outer SVG tag (but then we have to build the component wrapper, and the example above won't work anymore, so maybe not for V1)
- [ ] Adjust theo so that it only generates common.js for this token (do we ever need CSS version of these icons' paths?)
- [ ] Perhaps make the icon generator run automatically during pre-commit (not sure if we even want this, maybe not for V1)


OTKit/otkit-icons/token.yml
```diff
@@ -0,0 +1,27 @@
+props:
+  ic_information:
+    value: >-
```
karuto: Not sure where this `>-` coming from... need to debug 😭

matteofigus:
I like the idea thanks for exploring this @karuto.
Out of curiosity, do you have opinions about exporting only the paths VS the whole xml? I am trying to understand pros/cons of both.

ericz82:
Question: when a developer uses an icon token in a React project, will the syntax
```
render() {
    return (
      <div>
      {icons.ic_information} // here's the icon
      </div>
    );
  };
``` 
actually work or will `dangerouslySetInnerHTML` be required since the SVG raw is in XML syntax which would be escaped?

karuto:
@ericz82 It should just work, because the SVG markup is still in react land.

karuto:
@lastquestion 

> I wonder if we should be moving the work, logic, etc. from root into the actual otkit-icons. The script, etc. belong to the token repo, not the monorepo as a whole.

I'm ok with this, the location of the script (and the assets folder) is by no means finalized, is very portable.

> Should we actually save the original svg assets in this repository at all? shock.

Yeah I think that's what @matteofigus and @nickbalestra agreed upon, so that designers can easily add / modify. They won't be exposed to consumers of the token anyway.

> I would say that we should only store the original SVG assets here if design wants to use this repo AND that they use these SVG assets as the file they edit.

Yes that's the intention. Feel free to chime in @francesyun.

> If we do store the SVG here, then perhaps we don't need to check in the YML file at all? It should happen as part of pre-publish, but it doesn't need to be checked in.

Good idea, you are right that the yaml is really a middle step and we only consume the common.js output in the end, I haven't explored on messing around with theo build steps yet.

karuto:
@matteofigus

> Out of curiosity, do you have opinions about exporting only the paths VS the whole xml? I am trying to understand pros/cons of both.

Yes I do, there are pros and cons for each, and we can potentially support both. I will reply you tomorrow, it's pretty late here in UK 😭

lastquestion:
@karuto, @ericz82 : @ericz82 is correct. https://codepen.io/anon/pen/KQXyWa?editors=1111

karuto:
@lastquestion Thanks for the snippet, in that case I suggest we can set innerHTML in the icon component in ot-react-ui, so that end consumers won't need to worry about it and will just use the composited component directly.

francesyun:
> I would say that we should only store the original SVG assets here if design wants to use this repo AND that they use these SVG assets as the file they edit.
Yes that's the intention. Feel free to chime in @francesyun.

Yes that sounds right.

lastquestion:
Closed in favor of #135

PR #132, opened 2018-03-05
Bring typography token + little fix

matteofigus:


style-guide/src/styles/index.module.css
```diff
@@ -33,7 +32,7 @@ a {
 
 .nav-link {
   margin-bottom: spacing-medium;
-  font-weight: font-weight-medium;
+  font-weight: 500;
```
matteofigus: quick fix to get the styleguide published

karuto:
Wow, this fundamentally changes how desktop typography would work, I think we need to have a discussion about the namings because @mannionaco may want to introduce some level of context to the names so they are not just tshirt sizes.

matteofigus:
yeah let's change this - it was a quick one to be able to unblock the styleguide release

Issue #119, opened 2018-02-28
Proposal: remove package locks


matteofigus:
I am not a fan of having package locks in modules, I rather think they should be handled in the "end" app to lock the various subdeps. In that way, there are more opportunities to optimise (deduping) + keeping all the deps to the latest version (which is something now dependencies.io does for free).

Any objections on gitignoring it?

nickbalestra:
Agree couple of extra points in favor to this:
- at the moment this repo its been orchestrated with lerna without relying on its `--hoist` option or on similar alternatives like `yarn workspaces`, a package.lock  will simply lock devDeps in the root, so not that useful.
- furthermore tokens are pretty much dependency-free, so not much point at all neither into locking them

Lastly, I'll also suggest that along gitIgnoring it, we also config [lerna accordingly ](https://github.com/opencomponents/oc-template-react/blob/master/lerna.json#L10)

lastquestion:
I agree with these points. We should .gitignore and then update lerna. Here is a PR to do that: #138

PR #135, opened 2018-03-06
Add OTKit-icons token and icon generation pipeline

lastquestion:
This continues from #106, but I 

a) moved the SVGs and the build script inside `otkit-icons`
b) make `npm develop`, `npm build` work
c) prepare to actually extract the SVG paths, but for now, just do the entire SVG dumped out. 

I am unsure about:

* do we ever expect the format SCSS or CSSModule to be used for this icons? or only JS? 

I kind of want to only generate commonJS for this one, because I am having a hard time seeing how CSSModule or SCSS could use the SVG into CSS...

Now, I discovered that we can't really just remove the SVG root element unless we really have just one path as the SVG itself; and that even SVGO cannot optimize out the "def"/"xlink"s, nor combine certain kinds of paths. So we need to get design to make sure the SVGs themselves are single path for the "perfect flow" to work - that is when we extract out only the path, and allow the downstream platform specific clients to choose color, etc.

fyi: re: `package-lock.json` and issue #119 - I agree - this PR doesn't have package-lock checked in - but it's not gitignored. We should do that all together in a PR to address that issue.


OTKit/otkit-icons/ic_checkbox.svg
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg width="24px" height="24px" viewBox="0 0 24 24" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" style="background: #FFFFFF;">
+    <!-- Generator: Sketch 46.1 (44463) - http://www.bohemiancoding.com/sketch -->
+    <title>icon/ic_checkbox</title>
+    <desc>Created with Sketch.</desc>
+    <defs></defs>
```
lastquestion: this example doesn't actually use a def, but some icons we have right now do, and they require a `id` and `xlink`, which seems to conflict sometimes (see https://github.com/opentable/start-page-nodejs/pull/3313/) as an example

lastquestion:
If we're ok with the general approach here, I'll add in a page for the styleguide too.

mannionaco:
Hey Lin, all of the icons located here have been built in Sketch and reduced to a "combined shape"
screen shot 2018-03-06 at 10 10 18 am

if i open the "combined shape"
screen shot 2018-03-06 at 10 10 36 am

What is "id" and what is "xlink"?

The link (https://github.com/opentable/start-page-nodejs/pull/3313) example looks like a design process error (how annoying, probably my fault - sorry about that, will not happen moving forward) and the icons we have in the library will all have unique names and if they don't we should flag as error if there is way to do that.

lastquestion:
@mannionaco  - awesome, I've found the actual icons you've mentioned in the comment. I'll use icons from that set as example icons in this PR. Let me update the PR.

lastquestion:
OK, what I've done is taken about 30 of the icons from @mannionaco and placed them into the assets directory. I think we should consider whether to keep the original folder structure or ask designers to create files with the correct name.

The sample output will warn on some icons:

```
    warning: ic_checkbox.svg has at least one root grouped path, try to convert into just paths
    warning: ic_cuisine.svg has at least one root grouped path, try to convert into just paths
    warning: ic_parking_details.svg has at least one root grouped path, try to convert into just paths
    warning: ic_thumbs_up.svg has at least one root grouped path, try to convert into just paths
icons generated successfully
```

And this is what a snippet of `token.yml` looks like
```
ic_table_categories:
    value: >-
      <path fill-rule="evenodd" d="M17 9H7v2h10V9zm2 0v9a1 1 0 0 1-2 0v-5H7v5a1
      1 0 0 1-2 0V9H3a1 1 0 0 1-1-1V6a1 1 0 0 1 1-1h18a1 1 0 0 1 1 1v2a1 1 0 0
      1-1 1h-2z"/>
  ic_tag:
    value: >-
      <path fill-rule="evenodd" d="M12.153 3.575l3.274-.997a1.21 1.21 0 0 1
      1.198.32l4.682 4.68c.323.325.446.793.316 1.208l-1.044 3.333-9.298
      9.299c-.44.439-1.176.415-1.644-.053l-6.788-6.789c-.47-.468-.493-1.204-.054-1.644l9.358-9.357zm1.16
      1.668l-8.485 8.485 5.657 5.657
      8.486-8.486.707-2.12-4.243-4.243-2.121.707zm2.83 2.828a2 2 0 1 1-2.83
      2.828 2 2 0 0 1 2.83-2.828z"/>
  ic_thumbs_up:
    value: >-
      <g fill-rule="evenodd"><path stroke-width="2" d="M19.997 10.647A1 1 0 0 0
      19
      9.73h-4.503l.58-1.387c.568-1.354.568-2.343.09-3.059-.344-.517-.958-.934-1.878-1.235a1
      1 0 0 0-1.17.441l-3.7 6.24H5a1 1 0 0 0-1 1v7a1 1 0 0 0 1 1h12.36a1 1 0 0 0
      .981-.804l1.656-8.279z"/><path fill-rule="nonzero" d="M7 10h2v10H7z"/></g>
  ic_time:
    value: >-
      <path fill-rule="evenodd" d="M13 11h1.5a.5.5 0 0 1 .5.5v1a.5.5 0 0
      1-.5.5h-3a.5.5 0 0 1-.5-.5v-5a.5.5 0 0 1 .5-.5h1a.5.5 0 0 1 .5.5V11zm-1
      10a9 9 0 1 1 0-18 9 9 0 0 1 0 18zm0-2a7 7 0 1 0 0-14 7 7 0 0 0 0 14z"/>
```

As you can see some of the icons are simple paths now, some are groups. I cannot remove the `fill-rule` because - some of the paths require `evenodd` and some `nonzero`. So I think this is the best we can do.

nickbalestra:
LGTM as a first iteration

PR #138, opened 2018-03-08
remove package-lock.json

### LINKED ISSUE #119: Proposal: remove package locks

lastquestion:
Closes #119


matteofigus:
Thanks!

PR #141, opened 2018-03-08
[DX-449] Integrate icons into the styleguide

matteofigus:
screen shot 2018-03-08 at 15 20 55

Just a first go to see how it looks like /cc @lastquestion @francesyun @karuto @nickbalestra


francesyun:
lgtm

lastquestion:
I am going to manually run the ghpages publish because the last publish of `0.0.1` shouldn't have published this since it didn't exist yet.

Issue #142, opened 2018-03-09
Handling typography in responsive web


francesyun:
Brought up by @char1e5 (Charles works on works on RestRef, which is responsive down to mobile) 

What's the recommended strategy for bringing in typography tokens on responsive web? Because the token names for desktop-typography and mobile-typography appear to be the same (font-size-small, etc), when they are both imported, the names will collide. 

@mannionaco, is the idea that desktop-typography font stack will be used across the board for web (including responsive web at mobile size)? If so, this point is moot.

char1e5:
Great 👍 Thanks @francesyun 
Same questions. I just put under this PR - https://github.com/opentable/design-tokens/pull/97

lastquestion:
So I guess I have *two parts* to this discussion.

This is all my personal opinion which was largely formed today after trying to update `ot-react-ui` components to use the two new modules.  So please feel free to discuss and let me know what you think.

First, the two modules and consuming them.

* `otkit-typography-desktop` defines the font brand name, values for each font grouping (heading, body, subtext, etc.). Also, it defines the font fallback stack, which is really only for web, but that's alright - the only platform we support on desktop is web.

* `otkit-typography-mobile` also defines the font brand name, values for each font grouping -- but the actual pixel values may be different !! I also expect that eventually we will define a font fallback stack for mobile web as well, which will likely take on different values - as we may rely more broadly on device fonts. But in any case the high level point is the names of the groupings are the same, but may have different values. For example, `heading-xlarge-line-height` might be different.

When using them using CSS-modules, you can rename when importing so you don't have collisions, if you expect to use them in the same file: https://github.com/css-modules/postcss-icss-values#aliases. I tested this earlier this year when playing around with CSS modules, and if you're using the postcss+webpack build path it should work.


Second, writing shared components that need to rely on typography values - e.g. stuff in https://github.com/opentable/ot-react-ui-components/, basically. 

For these, I think that we may we need to build styling separately. One CSS module would import from `desktop`, the other from `mobile`. The built outputs would both be packaged in the NPM module. It's a little similar to node versus browser built outputs, I think. I'm not quite sure exactly how this would be picked up in the hosting app yet. But it should at least allow the flexibility for hosting apps to support responsive or not. To be honest, of course, I haven't completely thought through all the details here, but I will absolutely open an issue in `ot-react-ui` and link it from here later today.

karuto:
### Why do we split the tokens
[...]

PR #145, opened 2018-03-13
Update desktop typography based on latest symbol library

karuto:
### Overview
This PR updates the desktop typography token based on @mannionaco's changes in the latest symbol library, including the `xlarge` and `xxlarge` changes mentioned in #97, which I'm closing in favor of this PR.

This is a breaking change and would require major version update. @mannionaco and @francesyun have reached an agreement on the naming convention at least on the desktop web.

The changset may be confusing, I recommend using split mode or viewing the file directly.

- We are removing aliases from typography tokens, so that each of the platform has a holistic local source of truth that is clearly visible to designers. This allows us to evolve each platform faster without worrying about shared values and deltas. For more context behind this decision see https://github.com/opentable/design-tokens/issues/142#issuecomment-371864080.
- We are renaming the font groups to a clearer, semantic-free, strictly-size-based system. Before we have "semantic names" like "subheading-small", "subtext-xsmall" where we are mixing (very specific) semantics and values, trying to please both crowds - big mistake. During actual development this brought extra confusion and devs had to look up values behind names each time, which is the exact opposite of what these font group names should do.
 - Added comments following the designer's format in `otkit-colors` to improve maintainability.

@mannionaco updated reference:
otkit_typography_031318

### DIFF
--- a/OTKit/otkit-typography-desktop/token.yml
+++ b/OTKit/otkit-typography-desktop/token.yml
@@ -1,118 +1,143 @@
 props:
-  font-family:
-    type: "raw"
-    value: "'BrandonText', -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Helvetica, Arial, sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol'"
+  # The font name of OpenTable's desktop web brand font.
+  # =============================================
   font-family-brand:
     type: "raw"
     value: "BrandonText"
-  font-size-xsmall:
-    type: "size"
-    value: "{!font-size-xsmall}"
-  font-size-small:
-    type: "size"
-    value: "{!font-size-small}"
-  font-size-medium:
-    type: "size"
-    value: "{!font-size-medium}"
-  font-size-large:
-    type: "size"
-    value: "{!font-size-large}"
-  font-size-xlarge:
-    type: "size"
-    value: "{!font-size-xlarge}"
-  line-height-xsmall:
-    type: "size"
-    value: "{!line-height-xsmall}"
-  line-height-small:
-    type: "size"
-    value: "{!line-height-small}"
-  line-height-medium:
-    type: "size"
-    value: "{!line-height-medium}"
-  line-height-large:
-    type: "size"
-    value: "{!line-height-large}"
-  line-height-xlarge:
-    type: "size"
-    value: "{!line-height-xlarge}"
+
+  # Backward-compatible, generic font weights of OpenTable's desktop web brand font.
+  # Should only be used when composing font faces for the brand font.
+  # =============================================
   font-weight-normal:
     type: "raw"
     value: "normal"
   font-weight-medium:
-    type: "number"
+    type: "raw"
     value: "500"
   font-weight-bold:
     type: "raw"
     value: "bold"
 
-  heading-xlarge-font-size:
-    type: "size"
-    value: "{!font-size-xlarge}"
-  heading-xlarge-font-weight:
-    type: "raw"
-    value: "{!font-weight-bold}"
-  heading-xlarge-line-height:
-    type: "size"
-    value: "{!line-height-xlarge}"
 
-  heading-large-font-size:
-    type: "size"
-    value: "{!font-size-large}"
-  heading-large-font-weight:
+  # The entire font family definition for OpenTable's desktop web.
+  # Apply to the appropriate root element such as <html>, <body> or `reactroot`.
+  # =============================================
+  font-family:
     type: "raw"
-    value: "{!font-weight-bold}"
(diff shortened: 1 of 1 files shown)


OTKit/otkit-typography-desktop/token.yml
```diff
@@ -1,115 +1,143 @@
...
+
+  # Design decisions for font group sxmall-medium.
+  #
+  # =============================================
+  xsmall-medium-font-size:
     type: "size"
-    value: "{!font-size-xlarge}"
-  heading-xlarge-font-weight:
+    value: "12px"
+  xsmall-medium-font-weight:
```
karuto: We have decided to call the 500 weights "-medium" instead of "-semibold" as the screenshot indicated because of a few reasons:

- What designers really mean is they want `font-weight: 500`. In desktop web context, medium = 500 and semibold = 600, which may be confusing.
- Most font families matches medium to 500. The only difference is iOS, where @mannionaco wants to use semibold (600) to mean medium because San Francisco font is too skinny / thin at the 500 level comparing to all other fonts. Again, this is an anomaly not a standard.

francesyun:
LGTM, thank you vincent and kellen! 
The only thing that threw me off was "medium-bold" because the word "medium" is being used both as a font-size and font-weight. Can't think of a better option though.

OTKit/otkit-typography-desktop/token.yml
```diff
@@ -1,115 +1,143 @@
...
-    value: "{!font-weight-normal}"
-  subtext-xsmall-line-height:
+    value: "bold"
+  xxlarge-bold-line-height:
     type: "size"
-    value: "{!line-height-xsmall}"
+    value: "56px"
+
 
 imports:
```
matteofigus: shouldn't we remove the dependency from the aliases?

OTKit/otkit-typography-desktop/token.yml
```diff
@@ -1,115 +1,143 @@
...
+    value: "24px"
+  large-bold-font-weight:
     type: "raw"
-    value: "{!font-weight-normal}"
-  bodytext-small-line-height:
+    value: "bold"
+  large-bold-line-height:
     type: "size"
-    value: "{!line-height-small}"
+    value: "40px"
```
mannionaco: large-bold-line-height = 32px, not 40px

*my fault, error on provided reference*

OTKit/otkit-typography-desktop/token.yml
```diff
@@ -1,115 +1,143 @@
...
+    type: "size"
+    value: "16px"
+
+
+  # Design decisions for font group small-regular.
+  #
+  # =============================================
+  small-regular-font-size:
+    type: "size"
+    value: "14px"
```
mannionaco: small-regular = 16px

*my reference sheet is using iOS as an example*

OTKit/otkit-typography-desktop/token.yml
```diff
@@ -1,115 +1,143 @@
...
-  heading-large-font-size:
+
+  # Design decisions for font group xsmall-regular.
+  #
+  # =============================================
+  xsmall-regular-font-size:
     type: "size"
-    value: "{!font-size-large}"
-  heading-large-font-weight:
+    value: "12px"
```
mannionaco: xsmall-regular = 14px

*my reference sheet is using iOS as an example*

mannionaco:
Updated typography reference sheet:

otkit_typography_031318

OTKit/otkit-typography-desktop/token.yml
karuto: @matteofigus yes, I can remove it 😆

karuto:
@francesyun @mannionaco Thanks for taking a look! I have updated all the values based on the "desktop" section of your latest reference sheet, please check again.

lastquestion:
1. Did someone update the styleguide based on the reference that @mannionaco put up here?
2. This needs to be a major bump, and we need to react out and let everyone know to get this update pushed through. Does design - @mannionaco / @francesyun have a migration recommendation document?

matteofigus:
@lastquestion the styleguide doesn't look updated. After #146 is finalised, it will be done automatically by CI. In regards of updating the consumers, perhaps an e-mail to the OT web platform group could be a good idea? Would you be able to take care of it @karuto ?

lastquestion:
I think the email is a good idea, but the email should be actionable, so we need to give proactive advice to people - what names map to what. I think this depends on design. For example, what was `subheading` is now... what? `small-medium`?

mannionaco:
@lastquestion is there is place within github where I can log updates? I'd like to have a single source that we can go back to and reference in case someone doesn't see the email. Better yet, if this is a thing, I'd like to log the updates/changes and push/automate an email from one channel.

PR #146, opened 2018-03-15
[DX-428] Automatic-release 

nickbalestra:
The following to add automatic release script for CI


CONTRIBUTING.md
```diff
@@ -9,11 +9,9 @@ First of all, thank you for contributing. It’s appreciated.
...
 
-# To release new versions
+# Nothing is needed to release a new version, therefore **please don't manually publish tokens**
 
-1. To check what needs to be published run `npm run updated`.
-2. Make sure you are listed among the collaborators for the related packages that you want to release and that you are logged in with that  account to npm. To check if you are among the allowed collaborators of a specific token: `curl https://www.npmjs.com/package/<token-name>/collaborators` ([example](https://www.npmjs.com/package/otkit-typography/collaborators)). If you aren't just rechout to any of the person listed and ask to be added.
-3. To publish run `npm run publish` and follow the instruction on the terminal. The CLI will walk you through the publishing for each token that need to published.
+Upon merging to master a PR an automatic release system will make sure that related tokens that need to be updated get published on the public NPM registry following the rules listed below
 
 ## Version cheat sheet:
```
matteofigus: can we also rephrase this version cheat too. Instead of "release a patch..." perhaps something like "A patch will be released when...."

README.md
```diff
@@ -147,13 +147,12 @@ If you are using a token in your project, you can execute `npm link '<token-name
 
 ### Publish
 
-Since every token is a `npm` package, changes to the token, even if the code is merged, must need a new version published to take effect.
+Publish is automatic upon a PR get succesfully merged to master. **Please don't do manual publishing, in case of problem, open an issue**
+The following rules are automatically identified in CI:
```
matteofigus: I agree about a message in the readme to ensure nobody publishes, but perhaps, for everything else, I would link to the CONTRIBUTING's section? Perhaps with a link like "click here for more details..."?
What do you think

utils/automatic-release/process.js
```diff
@@ -0,0 +1,121 @@
...
+const differ = require('jest-diff');
+const diffToSemver = require('./diffToSemver');
+const console = require('console');
+
+// run lernaUpdate and return a list of the packages that need to be updated
+const root = path.join(__dirname, '../../');
+const latestTemp = path.join(__dirname, './tempNpm');
+
+const checkUPdated = () => {
+  const cmd = spawn('npm', ['run', 'updated'], {
```
matteofigus: I found you can do `lerna updated --json --loglevel=silent` for just getting this output in json. I think this could be a bit more robust?

Example:
```js
node node_modules/.bin/lerna updated --json --loglevel=silent
[
  {
    "name": "oc-template-react-compiler",
    "version": "5.0.0",
    "private": false
  },
  {
    "name": "react-app",
    "version": "1.0.19",
    "private": true
  }
]
```
In this way I think you can also easily skip the private ones here - which I think it would be nice?

utils/automatic-release/process.js
```diff
@@ -0,0 +1,121 @@
...
+
+const run = () => {
+  return new Promise((resolve, reject) => {
+    checkUPdated()
+      .then(updated => {
+        Promise.all(installLatestFromNPM(updated)).then(() => {
+          updated.forEach(pkg => {
+            const updatedPath = pkg.match(/^ottheme/)
+              ? `OTTheme/${pkg}/token.yml`
+              : `OTKit/${pkg}/token.yml`;
```
matteofigus: This name selector seems a bit specific to our use case. Can we move this path resolver at least to a separate file?
Perhaps some kind of `conf.js` like
```js
module.exports.pathResolver = pkg => ...
```
So that in the future this is a bit easier to spot/change/configure?

utils/automatic-release/process.js
```diff
@@ -0,0 +1,121 @@
...
+                console.log(`Error publishing ${pkg} as ${version}`, err);
+                throw err;
+              });
+            } else {
+              console.log(`Skipping publishing ${pkg}`);
+            }
+            console.log(diff);
+            console.log('--------------------------------');
+          });
+          fs.removeSync(latestTemp);
```
matteofigus: should this be in a finally instead? So that we are sure it's performed in any case

utils/automatic-release/test/diffToSemver.test.js
```diff
@@ -0,0 +1,37 @@
...
+});
+
+test("return 'major' when there is a breaking change on current keys", () => {
+  expect(
+    diffToSemver(
+      { color: '#000000', bg: '#000000' },
+      { supercolor: '#000000', bg: '#000000' }
+    )
+  ).toBe('major');
+});
```
matteofigus: Can we add another test with major being selected in case of
```js
{ color: '#000000', bg: '#000000' },
{ bg: '#000000' }
```
as the first test has addition and removal, and I think it would be nice to also test just the removal?

README.md
nickbalestra: Good idea!

utils/automatic-release/process.js
nickbalestra: OMG - love you!

utils/automatic-release/process.js
nickbalestra: Very good idea, I likes it, thanks for it!

utils/automatic-release/test/diffToSemver.test.js
nickbalestra: Sure think, good idea

utils/automatic-release/process.js
nickbalestra: Added in case of error as well, until `.finally()` is avail on node (currently on stage4) - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/finally

nickbalestra:
@matteofigus addressed all inputs, thanks for it!
Let me know if you have any other suggestion

utils/automatic-release/diffToSemver.js
```diff
@@ -0,0 +1,37 @@
...
+      version === undefined
+    ) {
+      version = 'patch';
+    }
+  }
+
+  return version;
+};
+
+module.exports = diffToVersion;
```
lastquestion: this is cool, cool. 👏 
Nice idea!!

utils/automatic-release/process.js
```diff
@@ -0,0 +1,125 @@
...
+const checkUPdated = () => {
+  const cmd = spawn(
+    'node',
+    ['node_modules/.bin/lerna', 'updated', '--json', '--loglevel=silen'],
+    {
+      cwd: root
+    }
+  );
+  let result = '';
+  cmd.stdout.on('data', function(data) {
```
lastquestion: :nitpick: how comes this isn't arrow syntax? ;)

nickbalestra:
Yes @matteofigus I've opened [DX-460] for this

Issue #144, opened 2018-03-09
Icons as Common.js: yay or nay?


matteofigus:
After discussing with @antwhite and many other stakeholders, I realised that having the icon exported as common.js is possibly less useful than I originally thought.

These are the reasons:
1) As they are now, the whole icons are available in the token, and when using them in react-land webpack is not gonna be able to do tree shaking. The way to fix it, should be to possibly hook into theo's generated file, and possibly produce individual icons. Example:
```js
import ic_star from `otkit-icons/icons/ic_star.js`;
```
2) If you have multiple icons in the page, having the `<svg>` as markup is less ideal for performance reasons. With css, you can have and reuse a css class multiple times in the page, while with JS a developer could make the mistake of repeating the markup multiple times.

3) @antwhite just made a benchmark with chrome to profile the difference between the 2 approaches and CSS seems to be reasonably faster anyways: https://github.com/opentable/ot-react-ui-components/pull/76#issuecomment-371869109

So, before we work on 1 (which we could totally do), I decided to open this issue to discuss: do we actually want common.js, or should we, once addressed the concerns about css modules (separate issue: #143) just distribute as CSS module? Perhaps this could be a chance to avoid people incurring on the possible issues and avoid doing some unnecessary work.

Vote 👍 or 👎 

/cc @antwhite @lastquestion @karuto @nickbalestra 

PS
If I am missing some valid reasons for distributing the common.js too, please add your comment 🙏

lastquestion:
I think we should distribute as commonjs *in the future* and only if design believes that we should have some set of icons which support *runtime variance*.

For example, if we have icons that should change color depending on user, or disable/enable state, things like this, that can be easily encoded as react components with props that affect the visual of the icon. I think this would be a call for desgn (cc @mannionaco @francesyun) to consider.

Once we do decide - if we decide - that we need it, then I would say we should definitely allow for treeshaking by distributing separate files, and we should only output the path. Given that, we will probably need to heavily customize theo for this specific design token, or just do it ourselves all at once directly in `build-icons.js`.

lastquestion:
This one is "yes, maybe in the future, not now" option. :)

nickbalestra:
Great questions Matteo!

[...]

PR #152, opened 2018-03-20
[DX-454] Don't distribute icons as common.js

matteofigus:
As discussed in #144 

Note: I am leaving the format generation as I can use it for the icons styleguide too (it uses npm link) - but I removed it from the publish, so that should be enough for now.


lastquestion:
Yes, so one thing is that it's difficult to programmatically iterate over CSS modules because we don't have access to metaprogramming. So although I agree with the spirit of the work I wonder if we should export some kind of JS array that simply names every icon. Again, I'm leading to consider that icons should be a specially generated token..

matteofigus:
For what do we need that @lastquestion ? 
In the case of the styleguide, the problem is solved as the thing is actually generated and npmignored but styleguide uses the npm link, so that is covered.

Do you need the full list for other reasons?

PR #155, opened 2018-03-20
[DX-457] Travis Setup for automatic-release

nickbalestra:
We still need to add the proper npm-token to the travis cofniguration. 
I've chatted with @MaximeDevalland and will pair for him to do that, once that is done we should be able to merge this and automatic-publishing should then be enabled.


nickbalestra:
I've added my npm token so for testing should be ok

PR #167, opened 2018-04-02
update to new colors

francesyun:
We have updated colors -- slightly bluer than before. Shouldn't break anything.


ericz82:
Can we choose whether we can to use lowercase or uppercase for hexadecimal colors and just stick with it all throughout?

karuto:
@ericz82 I suggest we leave that implementation detail to designers as it does not matter for developer's usage, CSS color hexes are case insensitive.

PR #168, opened 2018-04-03
Functional token for ticketing experience on desktop web

karuto:
### Overview
This PR introduces the first "functional token" (issue https://github.com/opentable/design-tokens/issues/91) that's intended to capture all the design decisions about a specific user experience segment / product: ticketing experience on desktop web. The design decisions originated from @francesyun. 😇 

In V1 it will only be used for ticketing modal that I developed. In time, we would like this to be shared across other desktop web ticketing experience products, such as booking flow and rest ref segments. 

The ticketing modal CSS is already 95% design tokens, therefore this token only captures design decisions that are NOT present in other universal tokens. Let me know what you think of this approach.

The variable names are self explanatory enough. I'm very flexible with the naming conventions of the token or the values!

@nickbalestra @matteofigus Let me know what your opinion is on integrating this into the style guide - or whether we should.

Preview:
image

### DIFF
--- a/OTKit/otkit-ticketing-desktop/token.yml
+++ b/OTKit/otkit-ticketing-desktop/token.yml
@@ -1,0 +1,22 @@
+props:
+  # Used in ticketing modal of desktop restaurant profile
+  # =============================================
+  modal-margin-vertical:
+    type: "size"
+    value: "48px"
+
+  # Used in ticketing modal of desktop restaurant profile
+  # =============================================
+  modal-max-width:
+    type: "size"
+    value: "672px"
+
+  # Used in ticketing modal of desktop restaurant profile
+  # =============================================
+  modal-hero-fallback-background-color:
+    value: "#492c6e"
+
+global:
+  platform: "desktop"
+  type: "ticketing"
+  category: "ticketing"
--- a/OTKit/otkit-ticketing-desktop/LICENSE
+++ b/OTKit/otkit-ticketing-desktop/LICENSE
@@ -1,0 +1,20 @@
+Copyright (c) 2017 OpenTable, Inc.
+
+Permission is hereby granted, free of charge, to any person obtaining
+a copy of this software and associated documentation files (the
+"Software"), to deal in the Software without restriction, including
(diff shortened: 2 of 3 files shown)


OTKit/otkit-ticketing-desktop/token.yml
```diff
@@ -0,0 +1,27 @@
...
+  # Used in ticketing modal of desktop restaurant profile
+  # =============================================
+  modal-max-width:
+    type: "size"
+    value: "672px"
+
+  # Used in ticketing modal of desktop restaurant profile
+  # =============================================
+  modal-background-color:
+    value: "rgba(0,0,0,0.8)"
```
karuto: Not sure if this should live in `otkit-colors` or not, we should move it to there if this is universal.

OTKit/otkit-ticketing-desktop/token.yml
```diff
@@ -0,0 +1,27 @@
...
+
+  # Used in ticketing modal of desktop restaurant profile
+  # =============================================
+  modal-hero-fallback-background-color:
+    value: "#492c6e"
```
karuto: I don't think @francesyun approved this, but I added a shade of purple that is close to the hamster ball hero image. Feel free to change @francesyun 😭

OTKit/otkit-ticketing-desktop/token.yml
francesyun: this value will be used exclusively for all modal backgrounds. is that enough to move it to otkit-colors?

OTKit/otkit-ticketing-desktop/token.yml
francesyun: haha, this is fine with me :)

OTKit/otkit-ticketing-desktop/token.yml
karuto: yes we should move it to colors, would you like to do it? Or I can do it later :P

karuto: Either way I am going to remove this from ticketing token for now

PR #169, opened 2018-04-03
Update README for documentation on how to use icons

ericz82:


lastquestion:
And, it might be broken anyway, because I'm not sure that @nickbalestra picked up the task to change the SVG string to include the full file, not just the path.

ericz82:
Huh? How are the icons working in the style guide then?

If this is not yet ready, would someone else be willing to update the README on how to consume icons as design tokens when the time comes?

lastquestion:
Well, the publish process seems broken, or odd. 
It seems that someone published `0.1.1` of icons, even though they didn't put the tags.
perhaps the auto build process still has bugs in it:
```
lxu (master u= origin/master) ~/src/design-tokens $ git tag
....
otkit-icons@0.1.0
...
lxu (master u= origin/master) ~/src/design-tokens $ npm show otkit-icons
{ name: 'otkit-icons',
  'dist-tags': { latest: '0.1.1' },
  versions: [ '0.1.0', '0.1.1' ],
```

So I'm not sure if the last publish has @matteofigus's change to drop commonjs. 
Also, I'm not sure if the current styleguide on https://opentable.github.io/design-tokens/ is up to date.
So it might be working because it hasn't been published yet.

lastquestion:
Oh, now I remember, styleguide works because of `npm link` which saves us as the file is still generated. I recall thinking about this problem with @matteofigus 
 in https://github.com/opentable/design-tokens/pull/152. In any case, yes, the README should explain how to use icons via CSS modules, because the way it's done in the styleguide is not usable by consumers via NPM.

nickbalestra:
@lastquestion not yet, wip

matteofigus:
@lastquestion @ericz82 sorry for this inconvenience. The issue seems related to the fact we are adding a github token of a user (nick's for now but we'll switch to the OT official one when we finish testing the beta) that has to be admin of the repo. In this case, the PR runs on top of eric's fork (https://github.com/ericz82/design-tokens) and that doesn't have that user as admin - therefore travis fails.

The solution (@nickbalestra is working on a PR now) is to run that command only on master/upstream ([here](https://github.com/opentable/design-tokens/blob/master/.travis.yml#L15)) as that's the only point where this is actually required.

matteofigus:
BTW this should go green now that we made #171. In any case as @lastquestion said common won't be supported, CSS modules will be supported as soon as @nickbalestra completes DX-453

karuto:
Closing this since we have implemented icon component with CSS modules now.

PR #173, opened 2018-04-04
^ on styleguide

matteofigus:


matteofigus:
@karuto no I think that's a separate issue - this is mostly as cleanup

PR #174, opened 2018-04-05
Adding additional font variations.

jwhyte88:
Adding these additional font variations following on from a catchup last week with @lizgersh and @miwa505 .
These fonts will be used for the JA version of the new reservation reconfirmation page.

### DIFF
--- a/OTKit/otkit-typography-desktop/token.yml
+++ b/OTKit/otkit-typography-desktop/token.yml
@@ -69,6 +69,20 @@ props:
     value: "24px"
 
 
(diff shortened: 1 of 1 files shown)


matteofigus:
@karuto thanks for the review, still shouldn't a designer review this kind of changes before merging? /cc @francesyun @mannionaco all good here according to you?

mannionaco:
These are not approved to go into design tokens.

mannionaco:
@matteofigus will do!

PR #178, opened 2018-04-09
[DX-473] Auto-publish fails when there is a new token (fix)

matteofigus:
1) When the token was unpublished, npm install would 404 and there would be a mess of events on the cmd.on('close') because of this: https://nodejs.org/api/child_process.html#child_process_event_close. More specifically *This is distinct from the 'exit' event, since multiple processes might share the same stdio streams*. The way I fixed it is by doing the same action is sequence.

2) There was also a "undefined" on the err diff, which I replaced with <DIFF NOT AVAILABLE> which is imho cleaner.

### DIFF
--- a/utils/automatic-release/lib/installPackages.js
+++ b/utils/automatic-release/lib/installPackages.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const fs = require('fs-extra');
+const Promise = require('bluebird');
 const spawn = require('cross-spawn');
 
 // locally install latest version of packages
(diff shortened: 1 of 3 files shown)


utils/automatic-release/lib/installPackages.js
```diff
@@ -1,21 +1,22 @@
...
+  return Promise.each(packages, pkg => {
     const cmd = spawn('npm', ['install', pkg], {
       stdio: 'inherit',
       cwd: latestTemp
     });
-    return new Promise(resolve => {
-      cmd.on('close', code => (code === 0 ? resolve(pkg) : resolve(null)));
-    });
+    return new Promise(resolve =>
+      cmd.on('close', code => resolve(code === 0 ? pkg : null))
```
nickbalestra: Noice :)

utils/automatic-release/lib/process.js
```diff
@@ -16,19 +16,23 @@ const latestTemp = path.join(__dirname, './tempNpm');
...
             if (version) {
               if (doPublish) {
-                console.log(diff);
-                console.log('--------------------------------');
+                printDiff();
                 return publishPackage(pkg, version, root);
               } else {
                 console.log(`TEST: "Should publish ${pkg} as ${version}"`);
-                console.log(diff);
-                console.log('--------------------------------');
```
nickbalestra: Thanks for this

PR #179, opened 2018-04-09
[DX-453] - Build Icons Utils / CSS-Icons

### LINKED ISSUE #143: Icons as CSS modules are broken
As we speak, icons (0.1.0) are available as css module in this format: 
```css
@value ic-payment: <path .../>;
...
```
So in short, we export the path only.
When using CSS modules, we can base64 or not:
```css
background-image: url("data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL...")
background-image: url("data:image/svg+xml;charset=UTF-8,%3Csvg xmlns='http://...'")
```

My proposal is:
1) We export the whole thing so that we can just do
```css
@value ic-payment from 'otkit-icons/token.cssmodules.css';
background-image: url(ic-payment);
```
2) We **don't** base64 encode because the full path could give us more opportunities to compress when this will ultimately be a css or js and be gzipped from the cdn when served via http/2

👍 if you agree?

/cc @lastquestion @karuto @antwhite @nickbalestra

nickbalestra:
This PR address few things
- [x] move the icon-build utility into the root of the repo together with the rest of the utils so that is not specific to otkit icons, but could be used by other icon-tokens. In general would be a good practice to keep devDependencies in the root manifest, and use the monorepo/lerna to orchestrate any specific thing related to dev/build/..
- [x] modularized and added test coverage for the icon-build utils
- [x] fixed #143. After some research I found this very [well written blog post](https://codepen.io/tigt/post/optimizing-svgs-in-data-uris) and decided to rely on `mini-svg-data-uri` for better results while addressing the issue.
- [x] update the style-guide to rely on the new icon-tokens 

This should publish a patch for the icons token and deploy the updated style-guide


utils/build-icons/test/generateTokenFile.test.js
```diff
@@ -0,0 +1,96 @@
+'use strict';
+
+jest.setTimeout(100000);
```
matteofigus: can we remove this?

utils/build-icons/test/getIconData.test.js
```diff
@@ -0,0 +1,61 @@
...
+    readFile: mockReadFile
+  };
+});
+
+// jest.mock("path", () => {
+//   return {
+//     extname: jest.fn(fileName => fileName.substring(fileName.indexOf("."))),
+//     basename: jest.fn((fileName, ext) => fileName.replace(ext, ""))
+//   };
+// });
```
matteofigus: remove comments

utils/build-icons/test/getIconFileList.test.js
```diff
@@ -0,0 +1,48 @@
...
+jest.mock('path', () => {
+  return {
+    extname: jest.fn(fileName => fileName.substring(fileName.indexOf('.'))),
+    basename: jest.fn((fileName, ext) => fileName.replace(ext, ''))
+  };
+});
+
+const getIconFileList = require('../lib/getIconFileList');
+
+test('', async () => {
```
matteofigus: can we give a name to this test?

PR #183, opened 2018-04-10
Simple pub

nickbalestra:
Moving away from lerna to handle publishing as we have some issues with it `--force-publish` command, while we keep relying on it for bootstrapping and getting a list of token via `lerna ls`.


nickbalestra:
Yes, they need updating

PR #187, opened 2018-04-12
Additional icons

mannionaco:
Adding icons to library that are missing


matteofigus:
@ericz82 PRs are often done by designers here in this repo without using the git CLI and we want them to own the workflow by lowering the entry barrier as much as we can.
In this perspective, I don't think we should ask them to squash commits - it's their repo.

ericz82:
@matteofigus I meant that the person who merges the PR can probably squash the commits. I worked with @mannionaco to get the icons to be added in the right format for design tokens. Sketch gave us some problems on the first two attempts which is why there is a total of 3 commits. The first two were wrong and should be ignored, which is why I suggest squashing.

ericz82:
@matteofigus Please feel free to merge this PR when ready. Our team will be using the new icons in our work very soon and would like to have our icons coming from design tokens now.

PR #194, opened 2018-04-19
Make dependencies.io styleguide PRs to resolve ^X.X.X versions

matteofigus:


matteofigus:
Apparently there is a feature for that but requires v2: https://www.dependencies.io/docs/configuration/

Hopefully it makes sense @davegaeddert

davegaeddert:
Hi @matteofigus -- I'm not sure I follow exactly, but I think you're on the right track. I'm guessing you'll want to add the same `constraint_prefix: ^` setting to the / config also.

davegaeddert:
Ok! And sorry, I just noticed the config isn't quite right -- `filters` should be under `manifest_updates`. You can take a look at the examples on that link you posted above. Hope that helps!

matteofigus:
Ok, addressed in https://github.com/opentable/design-tokens/pull/196

Issue #134, opened 2018-03-05
Styleguide generation bug


matteofigus:
Looks like when running `gatsby build` there is an error caused (probably) by otkit-typography css module outputting `  Error: Order in extracted chunk undefined`

This doesn't happen when we do `gatsby develop` and prevents the styleguide to be published.

PR #212, opened 2018-06-13
[breaking change] new colors; fix styleguide

lastquestion:
1. Implement the new colors, as per design. This adds colors that are named by content, not by intent. There is a mapping of old to new colors, which I will add to a comment below this PR description.
2. Group the colors by their base color.
3. Fix all the incorrect references to token values that were renamed, lost, changed, etc. In particular `typography` stuff was renamed and no one fixed the styleguide.

I want to figure out a way to throw errors when building for the variables thing, but I'll do that in a follow up PR.

### DIFF
--- a/OTKit/otkit-colors/token.yml
+++ b/OTKit/otkit-colors/token.yml
@@ -1,36 +1,201 @@
 props:
-  # Primary Action
-  # This color is only used on objects, backgrounds with large text (24px or larger, bold), and will be used as a color to guide users towards actionable content.
-  # =============================================
-  color-primary:
-    value: "#da3743"
-  color-primary-active:
-    value: "#b8222d"
-
-  # Primary Gray, used on all text sizes
-  # =============================================
-  color-gray-primary:
-    value: "#2d333f"
-
-  # Secondary Gray, used on smaller font sizes and for disabled primary buttons
-  # =============================================
-  color-gray-secondary:
-    value: "#6f737b"
-
-  # Utility Gray, used for borders, dividers, containers
-  # =============================================
-  color-gray-utility:
-    value: "#d8d9db"
-
-  # Lightest gray, used for backgrounds for the page and backgrounds for active/hover states
-  # =============================================
-  color-gray-active:
+  ash-lightest:
     value: "#f1f2f4"
 
-  # White, used for button text and (if necessary) backgrounds
-  # =============================================
-  color-white:
-    value: "#FFF"
+  ash-lighter:
+    value: "#d8d9db"
+
+  ash-light:
+    value: "#91949a"
+
+  ash:
+    value: "#6f737b"
+
+  ash-dark:
+    value: "#2d333f"
+
+  ash-darker:
+    value: "#141a26"
+
+  red-lightest:
+    value: "#fceeef"
+
+  red-lighter:
+    value: "#eea0a5"
+
+  red-light:
+    value: "#e15b64"
+
+  red:
+    value: "#da3743"
+
+  red-dark:
+    value: "#931b23"
+
+  red-darker:
+    value: "#450d10"
+
+  aqua-lightest:
+    value: "#eefcf9"
+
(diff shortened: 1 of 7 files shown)


lastquestion:
Colors look like this
screen shot 2018-06-12 at 5 27 09 pm

lastquestion:
Table of conversion of colors for this breaking change.

| Old color name       | New color name |
|----------------------|----------------|
| color-primary        |   red             |
| color-primary-active | red-dark               |
| color-gray-primary   | ash-dark               |
| color-gray-secondary | ash |
| color-gray-utility |		ash-lighter |
| color-gray-active |		ash-lightest |
| color-white |		white |

style-guide/src/components/section-header/styles.module.css
```diff
@@ -1,11 +1,12 @@
-@value spacing-large, spacing-medium, large-bold-font-size, large-bold-font-weight, large-bold-line-height, color-gray-primary, color-gray-utility from '../../styles/tokens.module.css';
+@value spacing-large, spacing-medium, large-bold-font-size, large-bold-font-weight, large-bold-line-height from '../../styles/tokens.module.css';
```
lastquestion: I would actually prefer to remove `tokens.module.css` and directly reference the token values from the NPM module names, but I decided to try and keep the diffs small.

OTKit/otkit-colors/token.yml
```diff
@@ -1,36 +1,201 @@
...
+  blue:
+    value: "#4a6fde"
+
+  blue-dark:
+    value: "#2146b5"
+
+  blue-darker:
+    value: "#0d1b45"
+
+  fuchia-lightest:
```
wesww: `fuchia` ->
`fuchsia`

lastquestion: LOL. Oh snap @mannionaco I think I pulled these by screenscraping the invision, you have to fix it there too ;)

wesww: it's a nitpick, but one that will bother me if it stay in and gets put into use everywhere

style-guide/src/components/section-header/styles.module.css
matteofigus: yeah this was an hack we did because of a gatsby issue, if I remember it uses an old version of the webpack extract plugin which doesn't support css modules to be compiled with ignoreOrder: true option and so this was enough for the styleguide to build without errors.
If now gatsby moved to webpack 3 or 4, perhaps the issue is gone and we can do this ;)

matteofigus: For your reference, issue and PR: https://github.com/opentable/design-tokens/issues/134

lastquestion: Oh, I didn't know! Hm. TIL. I did run `npm run build` in `styleguide` to do the build, but I didn't inspect the css nor verify the static site looked correct. It didn't fail to build, though. Let me check that tomorrow.

matteofigus: yeah, I mean this should be fine as you said in case it would be in a separate PR ;)

lastquestion:
I double checked and verified that `gatsby build && gatsby serve` work with the direct CSS module reference that I added.

OTKit/otkit-colors/token.yml
lastquestion: Fixed!

lastquestion:
We should send a email communication once we publish this to prepare downstream for the breaking change. A quick github search shows 32 hits in `package.json` for it. I'll take care of sending the email, unless someone else wants to do it (please please?)

lastquestion:
@wesww if you don't mind approving the PR I will merge and let's see if the build process automatically does a major version bump ;)

wesww:
looks good, don't let me slow you down

lastquestion:
:dance:

PR #216, opened 2018-06-20
remove ticketing token

### LINKED ISSUE #211: Why is ticketing in there?
Everything else in this repo is granular why do we have this? Feels like the antithesis to what design tokens is.
https://github.com/opentable/design-tokens/tree/master/OTKit/otkit-ticketing-desktop

lastquestion:
cc @karuto @johnnonolan @acolchado 
Closes #211


lastquestion:
Oh, that's a really good idea. ^_^ Let me fix that.

lastquestion:
Uh, funny thing, the thing was never mentioned in the README to begin with. I don't think it's worth adding something just to deprecate it. If you feel strongly about it, I don't mind adding it, but which section should I add it to? `otkit`?

matteofigus:
Oh right, I'll leave it to you so. One thing worth mentioning tho is that @karuto expressed some concerns on the issue so it may be worth double-checking everybody agrees about removing this.

Goodbye!

karuto:
I left a comment in #211 two weeks ago.

PR #221, opened 2018-07-05
[DX-512] Export colors as hex rather than rgb

matteofigus:
Is there is any particular reason for converting our colors from hex value (example `#ffffff`) to rgb (example `rgb(244, 66, 66)`)?

The reason is that TMS (our email system) is starting to use design tokens to render some emails, and rgb colors are not usable on old HTML attributes like `bgcolor="#fff"`. The team at the moment implemented a rgb2hex converter to solve the problem, but I think that if we have no particular reason, distributing hex values may just make things simpler?

If we would merge this, I would take care of auto-publishing this in a patch manner without breaking TMS /cc @opentable/unexpected-swallows


matteofigus:
Cool, thanks for the reviews. I'll merge and manually publish in a non-breaking way for TMS tomorrow.

karuto:
@matteofigus can you follow up on this please 😁

jwhyte88:
Thanks @matteofigus / @karuto / @lastquestion !

matteofigus:
@karuto done and published as patch. No existing consumer should be affected as far as I understand.

PR #226, opened 2018-07-13
fixing map and description view box to 24

mannionaco:


lastquestion:
Can we get another designer to approve? @francesyun

PR #225, opened 2018-07-13
adding metric change icons

mannionaco:
up and down

PR #227, opened 2018-07-13
updating icons per design request

mannionaco:


lastquestion:
Does this replace #225 ?

mannionaco:
Yes

lastquestion:
@mannionaco can we have another designer approve? maybe @francesyun

PR #229, opened 2018-07-18
support more complex SVG icons

lastquestion:
This will fail until the following issues are fixed:

```
ic_description.svg has a viewbox that isn't 24
ic_map.svg has a viewbox that isn't 24
there are duplicate ids:
        path-1
        mask-2
        Combined-Shape
        Oval-5
        icon/ic_favorite_filled
        ic_favorite_selected
        icon/ic_parking_details
        ic_parking_details
```

What I did here:

* remove much of the SVGO optimization passes, keeping on only things that are purely text & comment related
* change the error reporting to an accumulator model so we can collect all errors in every run
* remove various errors around single path, single group
* merge all ids across all icons and verify they are all unique; error if not
* generate full SVGs
* change the styleguide to generate inline SVG

I would probably like to refactor out the messages into a separate file but I'll do that as a followup.


lastquestion:
Needs #234 merged

lastquestion:
I fixed up the tests and now everything should pass.

lastquestion:
@karuto and @jrolfs can you review this?

Once you approve I am going to manually bump the package version to 4.x and publish locally, as the diffs in the actual SVG values will cause a minor change, not the major change we want.

style-guide/src/styles/otkit-icons.module.css
```diff
@@ -25,6 +26,11 @@
   justify-content: center;
 }
 
+.icon {
+  width: icon-size;
+  height: icon-size;
```
jrolfs: Can we just set one of the dimensions and rely on aspect ratio preservation? This'll kinda serve as a test and an example (does that make sense?).

utils/build-icons/lib/config.js
```diff
@@ -105,7 +105,7 @@ module.exports = {
         removeDimensions: true
       },
       {
-        removeAttrs: { attrs: '(stroke|fill)' }
+        removeAttrs: false
       }
     ]
   }
```
jrolfs: Why are we removing so much of the optimization? I know @aselbie mentioned he's seen it fail or mess with the intended vectors (definitely curious to hear a little more about that).

In the last SVG icon set I helped implement I believe we tuned it pretty similarly to how this had been configured and only whitelisted a couple attributes (`class` and `currentColor`?) used in targeting individual SVG nodes with CSS.

utils/build-icons/lib/processIcon.js
```diff
@@ -1,57 +1,24 @@
...
-      return console.log(messages.multiRootPaths(fileName));
-    }
-    const path = builder.buildObject({ g: innerContent.g[0] });
-    const viewBox = innerContent['$'].viewBox;
-    const xmlns = innerContent['$'].xmlns;
-
-    return { id, svgDataUri: buildSVGDataURI(xmlns, viewBox, path) };
-  } else {
-    return console.log(messages.missingTopLevelPathOrGroup(fileName));
-  }
```
jrolfs: I can't remember from when I actually spent some time with this stuff – are any of the current assets being handled by the other cases here? I feel like there were a couple.

Either way I thought it was pretty nice to have some of the edge cases handled. It might be nice to err on the side of caution ahead of time (in this case) so designers can throw a little more at it before needing to involve an engineer.

utils/build-icons/lib/config.js
jrolfs: Why are we removing so much of the optimization? I know @aselbie mentioned he's seen it fail or mess with the intended vectors (definitely curious to hear a little more about that).

In the last SVG icon set I helped implement I believe we tuned it pretty similarly to how this had been configured and only whitelisted a couple attributes (`class` and `currentColor`?) used in targeting individual SVG nodes with CSS. This worked great.

That being said I understand we may want to be a little cautious with optimizations that actually affect `path` data like `convertPathData`, `convertTransform`, `mergePaths`, and `convertShapeToPath`.

I'd be willing to do some more research/testing and follow up with a PR that re-enables some based on more information – but I still think we can trust more/most of these ;)

utils/build-icons/lib/config.js
```diff
@@ -105,7 +105,7 @@ module.exports = {
         removeDimensions: true
       },
       {
-        removeAttrs: { attrs: '(stroke|fill)' }
+        removeAttrs: false
```
jrolfs: I think we should still provide a whitelist here to establish convention  on how we actually target SVG nodes whether using CSS or JS.

It would also serve to remove clutter and potential selector mismatches though we should probably just be quite explicit with `className`.

style-guide/src/styles/otkit-icons.module.css
lastquestion: That would be interesting. Let me try it.

utils/build-icons/lib/processIcon.js
lastquestion: None of these "edge cases" should apply any more, because we now have moved id / path stuff to the designer's responsibility. In other words, we only used to care about this because we were trying to 

a) ultimately generate a single path instead of a full SVG
b) warn if there were ids because we were removing all of them

However, now we don't remove ids, and we instead require them to be unique. And since we are writing out the entire SVG, we don't care about multiple groups or paths.

utils/build-icons/lib/config.js
lastquestion: I don't see much value if we are sure to only send one copy of the used icon down per page. If we determine they are bigger we can re-enable some or we can get design to run something manually before committing it to the repo.

utils/build-icons/lib/config.js
lastquestion: Not sure what you mean; what's wrong with using the `id` as target?

utils/build-icons/lib/config.js
aselbie: Alright. So looking at this again, my experience with SVGO is mostly through https://jakearchibald.github.io/svgomg/. By default it has almost every setting on. If there are particular settings that you all are familiar with that have measurable benefits and limited downsides I'm completely cool with that. My experience with this particular tool is that it has broken a few icons and I've had to play with the settings (sometimes manually) until it didn't.

In general, though, me feeling is that we'll be adding icons at an asymptotically decreasing rate. I'm not sure that the effort/risk involved in automating (or adding it to the build pipeline in general) this optimization is worth the potential downsides.

jrolfs: > asymptotically decreasing rate

lastquestion: I tend to agree with @aselbie and also, we can change the optimization rules later in a follow up if needed.

lastquestion:
@jrolfs please re-review :)

utils/build-icons/lib/processIcon.js
jrolfs: Ah, okay. I think I misunderstood the purpose of the original code to some degree. 
Makes sense 👌

utils/build-icons/lib/config.js
jrolfs: My original comment was a little vague wrt "SVG nodes." I was referring to targeting nodes *within* an SVG, eg:

```svg
<svg>
  <symbol id="the-icon">
    <path d="[...]" />
    <path className="accent-color" d="[...]" />
    <path className="accent-color" d="[...]" />
  </symbol>
</svg>

<!-- [...] -->

<use class="prince" href="#the-icon"> 
```

```scss
// Default
#the-icon path.accent-color {
  stroke: $default-color;
  fill: $default-color;
}

// Prince's
#the-icon.prince path.accent-color {
  stroke: $prince-purple;
  fill: $prince-purple;
}
```

jrolfs: I guess really just the same reason we don't use IDs outside of SVG?

utils/build-icons/lib/config.js
jrolfs: > I'd be willing to do some more research/testing and follow up with a PR that re-enables some based on more information
= follow up

utils/build-icons/lib/config.js
lastquestion: What's wrong with
```css
#the-icon.prince #thehat {....} 
```

The reason why I think this might be worth doing is because it's trivial for designers to name paths as they wish via `id`, and we don't have to do any work to add `className` manually editing SVGs.

I think it works?

https://codepen.io/anon/pen/ejVZjE - note in this example I didn't use `<use>` but it should work as expected. Right?

lastquestion:
I am going to:

* disable the build pipeline
* merge to master
* manually publish at 4.0 from master,
* re-enable the build pipeline. 

Right now, the system believes it wants to do a patch release:

```
npm run automatic-release-test
...
TEST: "Should publish otkit-icons as patch"
```

and I am not quite sure how to make it think elsewise.

PR #233, opened 2018-07-18
Removing duplicates within paths in SVGs

mannionaco:
Note - Lightbulb has two paths, having issues converting that to a single path


lastquestion:
Two paths are OK :)

PR #234, opened 2018-07-18
removing path-1 and mask-2

mannionaco:
crossing my fingers...


lastquestion:
Merged, I am hoping this should bump a major version as the autopublish tool notices the changes that *weren't* picked up last build.

PR #236, opened 2018-07-19
fix styleguide build for v3 icons

lastquestion:
Fix the styleguide dependency with a major bump to get the styleguide to build properly so the automation is green. Prepares us to go to v4.


lastquestion:
OK, let's push this, it won't cause a build to go out but it will cause the styleguide to be published and up to date.

PR #242, opened 2018-07-31
only generate commonjs for otkit-icons for now

lastquestion:
cc @jrolfs @acolchado @karuto 

So, I was about to publish otkit-icons@4.0.0, when I discovered - by luckily hand running dry-run publish - that we are not publishing the commonjs.

Also, I notice that the cssmodule and SASS versions of the icons would have a giant full SVG XML markup, without enclosing `"` quotes.

For now, I would like to move forward by only outputting commonjs, and not the other versions for now. We can support the other versions in a patch followup by customizing the output to be a data-uri via adding custom code to the theo generation.

Right now, the build steps are still disabled in travis. We should merge this to master, then I can publish v4 and then push the tags and commit to master, and then reenable the build.

### DIFF
--- a/OTKit/otkit-icons/package.json
+++ b/OTKit/otkit-icons/package.json
@@ -18,8 +18,7 @@
   },
   "files": [
     "package.json",
-    "token.cssmodules.css",
-    "token.scss",
-    "token.yml"
+    "token.yml",
+    "token.common.js"
   ]
 }


jrolfs:
Maybe we want to rethink the build a bit? Make it a little less... trigger happy?

> We can support the other versions in a patch followup by customizing the output to be a data-uri via adding custom code to the theo generation.

https://github.com/jrolfs/design-tokens/blob/368ab519493f1d19fbb79b8bdd862da7e3aa7ede/utils/theo/transforms/svgDataUri.js#L1-L12
https://github.com/jrolfs/design-tokens/blob/368ab519493f1d19fbb79b8bdd862da7e3aa7ede/utils/theo/transforms/index.js
https://github.com/jrolfs/design-tokens/blob/368ab519493f1d19fbb79b8bdd862da7e3aa7ede/utils/theo/setup/icons.js#L1-L8

lastquestion:
@jrolfs I don't have any problems with the build as is, if you have suggestions for improving it or making it work better, all ears.

also, regarding the changes for cssmodules and SASS, I don't see how your proposed changes work, because that theo transform applies to all targets, as far as I am aware.

PR #246, opened 2018-08-01
adding more icons for restaurant side

mannionaco:


mannionaco:
@francesyun sorry i was wrong it didn't update the viewbox as I expected it to

OTKit/otkit-icons/ic_party_count.svg
```diff
@@ -0,0 +1,11 @@
...
...8092,15.2868877 L4.95281574,15.2868904 C4.70769063,15.2868904 4.46465381,15.2418286 4.23582208,15.1539516 C3.2046777,14.7579669 2.68977901,13.6010502 3.08575512,12.5699026 C3.82069874,10.6561088 5.1320025,9.5117912 7.0196664,9.13694976 C5.82934918,8.60434475 5,7.4096537 5,6.02125777 L5,5.91165437 C5,4.02744969 6.52744969,2.5 8.41165437,2.5 C9.71370481,2.5 10.8453945,3.22940104 11.4204073,4.30188698 C11.0517254,4.51502578 10.7078278,4.76626177 10.3940966,5.05021297 C10.0643294,4.36191626 9.36117108,3.88658243 8.54699262,3.88658243 C7.41646981,3.88658243 6.5,4.80305224 6.5,5.93357505 L6.5,5.99933709 C6.5,7.1298599 7.41646981,8.04632971 8.54699262,8.04632971 C8.56352789,8.04632971 8.58001737,8.04613366 8.59645886,8.04574374 C8.48054072,8.51136745 8.41901325,8.99849847 8.41901325,9.5 C8.41901325,9.67501296 8.42650641,9.84827579 8.44118941,10.0194852 C6.51730233,10.1578931 5.25085251,11.0336588 4.64183995,12.6467822 L4.64182597,12.646777 C4.44675453,13.1634729 4.70748271,13.7404744 5.22417865,13.9355458 C5.33706205,13.9781634 5.45672639,14 5.57738674,14 L7.95514558,13.9999938 C7.52906888,14.3735851 7.15548041,14.8065681 6.84758092,15.2868877 Z" id="Combined-Shape" fill="#2D333F"></path>
```
ot-raypeters: Use `ctrl + f` for `Combined-Shape` here... is this correct?

Looks like the build is failing because both of these have the same ID

mannionaco: Ok I found the issue. Will update.

lastquestion:
So do we need these filled icons now that we can just directly target the path for the unfilled versions and then fill it via CSS?

PR #279, opened 2018-09-25
adding icon for check indeterminate state

mannionaco:
@aselbie added the icon to the repo


aselbie:
Thank you. This fits the bill icon-wise. Should we maybe call it `ic_checkbox_indeterminate`?

PR #292, opened 2018-10-03
pin svgo to 1.0.4 (1.1.0 changed stuff)

lastquestion:
Some changes in svgo `1.1.x` changed the generated optimized SVG files and broke the tests.

Given the amount of work happening in that project let's pin directly to a version and update it manually via `dependencies.io`.


lastquestion:
cc @karuto

PR #307, opened 2018-10-16
updating xlarge and large

mannionaco:
xlarge 32px to 28px, line-height 40px to 36px
large 24px to 22px, line-height 32px to 28px

### DIFF
--- a/OTKit/otkit-typography-desktop/token.yml
+++ b/OTKit/otkit-typography-desktop/token.yml
@@ -126,13 +126,13 @@ props:
   # =============================================
   large-bold-font-size:
     type: "size"
-    value: "24px"
+    value: "22px"
   large-bold-font-weight:
     type: "raw"
     value: "bold"
   large-bold-line-height:
     type: "size"
-    value: "32px"
+    value: "28px"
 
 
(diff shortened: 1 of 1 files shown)


lastquestion:
@karuto can you take care of merging this as a major release? thanks!

karuto:
@lastquestion what do you mean by "merging this as a major release", aren't the version updates done by devops bot automatically now?

lastquestion:
the build script would make this a patch change, because we only changed the value. but I think this is a breaking change so it should be a major change.

karuto:
@lastquestion I agree that this is a major change in theory, but not sure about the execution. Should we manually change the major version in package.json?

lastquestion:
that would be insufficient. i would probably merge this PR into a local master, publish from there, and push that master up to remote master. the build script will still run, but since you already published there will be no changes.

karuto:
@lastquestion This commit has been manually merged into [master.](https://github.com/opentable/design-tokens/commits/master)

[A new release](https://github.com/opentable/design-tokens/releases) is also created.

Changes also published to [npmjs registry.](https://www.npmjs.com/package/otkit-typography-desktop)

PR #308, opened 2018-10-17
updating icon name

mannionaco:
adding checkbox to the indeterminate since they are in the same family


lastquestion:
It “should” do a major as a rename. Could pull this branch and test it manually and see what it proposes it will do

PR #330, opened 2018-11-09
adding service bell icon

mannionaco:


lastquestion:
@karuto is taking care of this.

PR #334, opened 2018-11-13
adding social icons

mannionaco:
the colors of the social icons are specific to the social networks brand. is that an issue with tokens?


mannionaco:
@karuto looks like this failed with "The command "npm test" exited with 1" is that an issue from my end?

mannionaco:
@pattilo got it to pass, had to update the path id

PR #367, opened 2018-12-19
updating shadow

mannionaco:


mannionaco:
@yuebwang created a PR for this change.

mannionaco:
@char1e5 and I have discussed the use case. This is an insignificant visual change that aligns to the styling within designer tooling for 1-to-1 mapping between design files and handoff to dev. The expected result is reduced confusion.

PR #376, opened 2018-12-27
modifying name

mannionaco:
from ic_indeterminate to ic_checkbox_indeterminate


karuto:
@mannionaco is this legit?

PR #398, opened 2019-01-18
adding icons for restaurant team

mannionaco:


karuto:
@mannionaco Don't worry, that's just a process error, no real error was detected on the changes you made. As you can see below that Travis runs again soon afterwards and it's green. I will merge it.

PR #410, opened 2019-01-29
feat(scss icons): GO-3673 Export otkit-icons as scss format

ot-raypeters:


ot-raypeters:
@lastquestion Here are the changes we discussed in slack. Please review async

PR #416, opened 2019-02-06
removing unused icon

mannionaco:
deprecating old version, this will be replaced in use by "ic-champagne-glasses"


mannionaco:
Closing in favor of #417

PR #420, opened 2019-02-12
Create OTKit token for grids

karuto:
### Overview
This PR creates the OTKit design token for grids, specifically for gutters and margins. Names TBD, awaiting @mannionaco's input.

### References
1024+
image
768 - 1023
image
320 - 767
image


OTKit/otkit-breakpoints/token.yml
```diff
@@ -10,5 +10,5 @@ props:
 
 global:
   platform: "core"
-  type: "breakpoints"
-  category: "breakpoints"
+  type: "breakpoint"
```
karuto: Changed this because all other tokens have singular form in this field.

OTKit/otkit-grids/token.yml
```diff
@@ -0,0 +1,20 @@
+props:
+  # These grid values are for viewports between 768px and 1023px.
```
jrolfs: Are we missing an intermediate set of values?

jrolfs: The screenshots above suggest that we are *not* missing them... however I don't understand why we don't just document this as `< 768` and `≥ 768`?

karuto: Because when I write notes, I want to mirror what design have. While what you said is true, it isn't how design spec view things. Design spec has 3 viewport ranges, so we document them as is, not our interpretation of it.

mannionaco: @jrolfs I @mentioned you in Abstract where these assets were pulled from. It should bring some clarity to why what is what.

karuto: @mannionaco would you mind approving this PR if you feel like you've addressed the concerns about the discussion above?

Issue #255, opened 2018-08-15
s/fuschia/fuchsia


mattiaerre:
maybe it is something you've already noticed but would it be possible to fix `fuschia` w/ `fucsia` or there are reasons why we are not doing it?

https://github.com/opentable/design-tokens/search?q=fuschia&unscoped_q=fuschia

lastquestion:
I think the misspelling slipped through. I am OK with a major bump

lastquestion:
Wait, isn't `fuschia` the corrct spelling?!

lastquestion:
https://en.wikipedia.org/wiki/Fuchsia_(color)

mattiaerre:
Yes you are right it is `fuchsia` but for sure NOT `fuschia` 😅

mattiaerre:
ping @lastquestion

karuto:
@mattiaerre sorry for the delay, but we will resolve this in the PR above. Closing this issue.

PR #421, opened 2019-02-12
Update fuchsia typo

karuto:
As #255 pointed out, we spelled "fuchsia" wrong for the longest time. This corrects it, but I would like to make sure that we spell it right in the design tools as well. @mannionaco 

This will be a breaking change, requiring a major version update.

### DIFF
--- a/OTKit/otkit-colors/token.yml
+++ b/OTKit/otkit-colors/token.yml
@@ -53,22 +53,22 @@ props:
   blue-darker:
     value: "#0d1b45"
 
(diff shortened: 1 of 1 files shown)


mannionaco:
@karuto will update in any documentation I have. Within design tools designers only see hex values.

karuto:
@mannionaco thanks for the feedback, that's glad to hear. Want to make sure we correct this typo once and for all.

karuto:
@mannionaco can you approve this PR in your review please?

PR #430, opened 2019-03-02
remove emojis

yuebwang:


lastquestion:
What's wrong with emojis? 😢

PR #435, opened 2019-03-08
Automatically generate SVG Icons for every color in OTKit Colors:

cbeley:
* Generated SVG’s are outputted to publicDist
     * Design tokens get created for each color variant and are now a url to where it would be in the CDN.
     * General code clean-up
     * Eslint clean-up


.eslintrc
```diff
@@ -0,0 +1,21 @@
+{
```
cbeley: Note that I choose a set of rules that mostly matched the code already.

cbeley:
Tests are failing because the pre-existing tests are spitting out every single design tokens: "The job exceeded the maximum log length, and has been terminated.".  I'll make some small modifications to have our build tests be less verbose.

utils/build-icons/lib/writeToken.js
```diff
@@ -8,6 +8,8 @@ const _ = require('lodash');
 const { messages } = require('./processIconHelpers');
 const { iconSize } = require('./config');
 
+const BASE_CDN = 'https://cdn.otstatic.com/otKitSvgs/';
```
chriscartlidge: Could we have this as an environment variable or configuration?

utils/automatic-release/test/process.test.js
```diff
@@ -108,8 +108,6 @@ jest.mock('../lib/versionAddCommitTagPackage', () =>
   jest.fn(() => Promise.resolve())
 );
 
-const versionAddCommitTagPackage = require('../lib/publishPackage');
```
karuto: We don't want to let jest mock this file anymore? why?

utils/build-icons/lib/writeToken.js
```diff
@@ -17,7 +18,7 @@ const writeToken = icons => {
         accum.errors = accum.errors.concat(val.error);
       }
 
-      accum.output[val.id] = { value: val.svg };
+      accum.output[val.id] = { value: url.resolve(process.env.DESIGN_TOKENS_CDN_PATH, val.transformedFileName) };
```
karuto: 👍

mannionaco:
@karuto do you know how the color variants being determined and for what benefit? I can provide design feedback based off how dev teams want to consume SVGs + Colors.

utils/automatic-release/test/process.test.js
cbeley: It's still mocked in the line above; however, versionAddCommitTagPackage is unused as a variable.  This was caught via linting.

cbeley: There's certainly more clean-up we can do with the mocks though.  They are all globally scoped, which caused some problems, but would take more time to fix than what I did in this PR.

cbeley:
As discussed, I changed this up to only generate svg icon color variants based on the values set in icon-colors.json.  I also added a README in otkit-icons talking about how this works/what it is.

OTKit/otkit-icons/README.md
```diff
@@ -0,0 +1,4 @@
+# otkit-icons
+All SVG's in this folder will be uploaded to https://cdn.otstatic.com/design-tokens/icons upon a new package publish, with each icon token set to its corresponding CDN url.
```
lastquestion: interesting choice of path. I like it. As a follow up, could we move the brandon text fonts into this same root path, and then we can actually have the CDN paths available as a design token, instead of copying and pasting everywhere?

OTKit/otkit-icons/README.md
```diff
@@ -0,0 +1,4 @@
...
+
+For any icon that has only a single fill color, multiple icons (and design tokens) will be generated automatically based on the colors set in `icon-colors.json`.  For example, if you have the icon `foo_icon` and `icon-colors.json` has `ash-dark`, a icon called `foo_icon_ash-dark` will be created automatically.  All colors in `icon-colors.json` must correspond to colors in `otkit-colors`.
```
lastquestion: What happens if an icon has multiple fill colors? Do we allow this? Do we error out during build, do we skip it, if so, do we log / write an message, warn, etc.

cbeley: We have two cases for this currently.  They just get ignored for the sake of generating alternate color versions of the icon (the base icon still gets generated though).

OTKit/otkit-icons/README.md
cbeley: Seems like a reasonable idea to me.  That's what I was thinking about going for /design-tokens, since I figured we may want to expand this in the future to other assets we want across consumer-web.

cbeley:
> We still need upload code (S3 bucket), which I suppose we need to decide if we want to hold that in this repo or do some private/public repo mirroring. I agree that should be in the next PR. This PR looks good. Thanks for cleaning up a whole bunch of stuff here.

Unfortunately, we can't merge this until our key requests goes through.  We've already chatted with ops about this and they are on-board with it all.  I'll ping you the ticket with full details about it all privately.

> However, I just had a realization as I was writing this - the styleguide will be broken once we merge this, right? Oh, I think the styleguide has some special secret magic where it directly reads the SVGs from the .js format output of otkit-icons (which is .npmignored). So maybe it still works. Can you just verify either way? I'm fine with the styleguide being broken in between PRs, because adding more to this PR is just going to complicate things.

It'll be broken until we publish the assets, but the change should otherwise be transparent to people when they bump versions.  Previously, we used data urls for the svgs.  Now, it's an actual url.  The original base icons with no -color-variant postfix are all still available..

PR #439, opened 2019-04-23
New icon that separates 'dot' from 'note pad'

benaz23:
Separated the dot from the path/symbol so the dot can be a different color to the rest of the icon.


ronniechong:
Looks good to me

PR #440, opened 2019-04-25
add japanese language font statck

lastquestion:
cc @Kimtaro 

Let's add the japanese language font stack to the design tokens for desktop typography so we can use them everywhere and don't need to copy and paste.


Kimtaro:
Like do we have a central location where we define the `BrandonText` `@font-face`?

lastquestion:
No, we don't. Everyone defines it themselves in various files. It's because we can't actually export CSS files in this repository. It was also hard because not everyone was using webpack / same configurations / etc. etc. 

However, I claim the problem should be shrinking because we are moving to one frontend infra / build system, e.g. CFE. 

Hypothetically, I should have also updated the styleguide with the correct CSS rules, and then we could just probably link to the styleguide CSS rules as what you should copy and paste into your repo. But the styleguide is not getting love either, so...

lastquestion:
I'm going to merge this now. this should cause a patch version to publish automatically.

PR #441, opened 2019-04-26
Adding spacing unit

mannionaco:
Adding 48px as xlarge spacing unit, changing 64px from xlarge to xxlarge

### DIFF
--- a/OTKit/otkit-spacing/token.yml
+++ b/OTKit/otkit-spacing/token.yml
@@ -8,6 +8,8 @@ props:
   spacing-large:
     value: "32px"
   spacing-xlarge:
+    value: "48px"
+  spacing-xxlarge:
     value: "64px"
 
 global:


mannionaco:
@karuto the restaurant design team will be adding 48px as a common spacing unit. I'm not sure if this is the correct path, but what I do know is this is probably a major change since I'm adjusting the naming. What do you think is the best approach to move forward?

karuto:
@mannionaco I don't think this will be a major change, adding a new value should be minor I think. So we can proceed as usual. I've approved. 😄

lastquestion:
@karuto I disagree. This should have been a major change, because we're changing the intent - I assume - of a spacing, and adding a new spacing to address the changed intent.

In other words, people who wrote `xlarge` should have been forced to change to `xxlarge`. 

As a minor change, this will break anyone who was using `xlarge` and expecting the semantic meaning of `64px`. 

There are 150 hits of `spacing-xlarge` in our codebases.

karuto:
@lastquestion I see what you mean now. That makes sense, I think I was making the judgement too literally based on the change-set alone. I think I can revert this PR, stop the auto publishing tools (because it'll try to publish a major - since we are removing a value), and manually publish a patch to get us back to the previous state. What do you think?

lastquestion:
Hmmm, I am not sure. We can also just let everyone know what happened and the recommended action. As it's just a search and replace it's not too expensive, because we didn't delete semantic meaning, just rename it.

Unfortunately (or fortunately) since this package is published in public NPM we can deprecate but not delete. So, we could also do as you suggest, revert, publish a new minor patch, and then do a major release.

Both approaches are likely OK. 64->48 means we lose 24px, probably most layouts will survive with that, so it probably won't destroy any layouts. On the other hand that's kind of a consumer response, restaurant uses design-tokens too and they're much more sensitive about the layout.

@mannionaco 's gut feel that this is major was right ;)

I guess after typing all this I've convinced myself that we should do as you suggest.

karuto:
@lastquestion Thanks for the feedback Lin.

To estimate the impact of this incorrect publish, I code searched the entire organization codebase, and only the following projects are using typography major version 3:
```
gc-reporting-message-composer-templates
gc-microapp-shift-notes
thunder-dash
photos-gallery-component
reservation-confirmation
oc-rbk
oc-dtp-react 
oc-reviews-widgets
```
Only 2 projects among them are using `spacing-xlarge`, the one we changed：
```
gc-reporting-message-composer-templates
gc-microapp-shift-notes
```
There is only one usage of `spacing-xlarge` in each project, used as a gutter / padding. 

Perhaps I can just ping the team and let them know that the next time they deploy, they should update otherwise they will have a smaller padding (-24px)?

PR #447, opened 2019-08-30
Add files via upload

benaz23:

PR #446, opened 2019-08-30
Add files via upload

benaz23:

PR #448, opened 2019-09-04
feat(otkit-icons): adds keyboard, waitlist and waitlist notification

ronniechong:
This PR is to add three icons into otkit-icons
- `ic-keyboard`
- `ic-reservation-waitlist`
- `ic-reservation-waitlist-notification`

icons

This is to consolidate these PRs:
https://github.com/opentable/design-tokens/pull/445
https://github.com/opentable/design-tokens/pull/446
https://github.com/opentable/design-tokens/pull/447

@benaz23


OTKit/otkit-icons/ic_keyboard.svg
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg width="24px" height="24px" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+  <g fill="#2D333F" fill-rule="nonzero">
```
Neha3011: @ronniechong I see that the fill is either none or currentcolor. Or these icons require specific colors?

ronniechong: This is a good point. A while back they have defined colours in those icons. Would love to know if they have changed the rules. Will update appropriately after Vincent shares his views

karuto: @ronniechong @neha-ot my view is that this `fill-rule` doesn't matter. Keep it as `nonzero` like the other icons do. If you need color filling, it should be defined at the `path` level not here. I'll approve this.

PR #480, opened 2020-05-11
feat(otkit-icons): add ic_bell icon

bronwyndunn:


lastquestion:
@char1e5 do we need bot for conventional commits to `otkit`? something like https://github.com/semantic-release/semantic-release ? its very nice DX.

PR #503, opened 2020-09-10
fix(styleguide): adds searchable icon filter

ronniechong:
This PR is adding a searchable filter for icons. The themable icon component is also refactored to use `useState`

filter


style-guide/src/components/design-nav/design-nav.js
```diff
@@ -2,9 +2,13 @@ import React from 'react';
 import { Link } from 'gatsby';
 import styles from '../../styles/index.module.scss';
 
-const NavLink = (props) => (
+const NavLink = props => (
   <div className={styles.navLink}>
-    <Link to={props.to} exact={props.exact} activeClassName={styles.linkActive}>
```
ronniechong, line 7: Gatsby 2 doesn't have `exact` props anymore

style-guide/src/components/design-nav/design-nav.js
```diff
@@ -2,9 +2,13 @@ import React from 'react';
...
+    <Link
+      to={props.to}
+      partiallyActive={props.exact ? false : true}
```
ronniechong, line 9: Using `exact` props to check for exact path. If we don't, refreshing the page will make the active link disappear.

style-guide/src/components/design-nav/design-nav.js
```diff
@@ -13,15 +20,15 @@ const NavLink = (props) => (
...
+    <NavLink to="/">All Design Tokens</NavLink>
+    <NavLink to="/otkit/colors/">Colors</NavLink>
+    <NavLink to="/otkit/typography/">Typography</NavLink>
+    <NavLink to="/otkit/grids/">Grids</NavLink>
+    <NavLink to="/otkit/spacing/">Spacing</NavLink>
+    <NavLink to="/otkit/borders/">Borders</NavLink>
+    <NavLink to="/otkit/breakpoints/">Breakpoints</NavLink>
+    <NavLink to="/otkit/shadows/">Shadows</NavLink>
+    <NavLink to="/otkit/icons/">Icons</NavLink>
+    <NavLink to="/otkit/icons-theme/">Icons (theme)</NavLink>
```
ronniechong, line 28: If there is no trailing slashes and if we refresh the page, the slash will automatically append to it and we won't be able to match the path and the styling will not apply.

We can use this if we want to keep non-trailing slash https://www.gatsbyjs.com/plugins/gatsby-plugin-remove-trailing-slashes/

style-guide/src/styles/otkit-icons.module.scss
```diff
@@ -49,3 +49,12 @@ $width-main: 768px;
   padding: $spacing-xsmall;
   margin: $spacing-small 0;
 }
+
+.form {
```
ronniechong, line 53: Since we are revamping, I'll leave this styling simple for now

jakeclements (approved):
V.cool, thanks @ronniechong!

PR #506, opened 2020-09-14
feat(style-guide): Resource box styling

neha-ot:


style-guide/src/components/handover-process/ResourceBox.js
```diff
@@ -4,12 +4,12 @@ import copy from '../../../static/handover-process';
 import styles from '../../styles/handover.module.scss';
 
 export default ({ resource }) => (
-  <div className={styles.resource}>
+  <div className={styles.resource} onClick={() => { window.open(copy[`${resource}.link`], '_blank')}}>
```
ronniechong, line 7: Semantically, `<a>` is more suitable for this as it is an actionable element

jakeclements, line 7: Can this be either an anchor tag or a button? It'd be more semantically correct than having a click handler on a presentational element

jakeclements (approved):
Thanks Neha

ronniechong (approved):
There!

PR #511, opened 2020-09-15
chore(build): removes dependencies.io config file

ronniechong:
This PR is removing the dependencies.io config for the time being until we find an alternative  tooling for package dependency automation.


jakeclements:
What's the reasoning behind removing dependencies.io?

ronniechong:
@jakeclements Originally I wanted to disable it until I can clean up the old dependencies PR. And I also thought if dependabot is part of github, why not use that instead relying another third party addon

PR #512, opened 2020-09-15
chore(build): Updates dependencies and reduce double styleguide-build

ronniechong:
This PR is patching all levels of vulnerabilities on the packages used.

Note: `optimist` package (deprecated package) is added here as a `theo` dependency for the time being until `theo` has merged the fix https://github.com/salesforce-ux/theo/pull/217

The `build-styleguide` in the automatic release is also removed since the travis script is already running it once.


jakeclements:
Does this raise any issues or changes to the way we work with this repo?

ronniechong:
@jakeclements  from my initial test run, there's no breaking changes on our current process

jakeclements:
Ancient one Ronz!

PR #518, opened 2021-03-31
[GRO-742] Add takeout icon

cosmonot1:


cosmonot1:
Icon already exists as `ic_pickup`. Closing PR

PR #523, opened 2021-06-14
Update px to rem

dethstrobe:


ot-raypeters (approved):
lgtm, but is this change going to cause issues in other apps?

dethstrobe:
> lgtm, but will moving to a relative unit of measurement cause issues for existing components when merged?

Not unless people are overriding the base html font-size. But we do want to do this, for the purpose of creating a responsive UI and allow for a more accessible site.

ot-raypeters:
> Not unless people are overriding the base html font-size. But we do want to do this, for the purpose of creating a responsive UI and allow for a more accessible site.

I agree on the benefits for responsiveness and accessibility.

Do we know of any apps overriding the base html font size?

dethstrobe:
### css overriding html font-size

- https://github.com/opentable/CHARM/blob/9b567de0963fd81670772508a234837cdc884540/FrontEnd/styles/charm_org.css#L6
- https://github.com/opentable/CHARM/blob/9b567de0963fd81670772508a234837cdc884540/FrontEnd/styles/charm_warm.css#L6
- https://github.com/opentable/LegacyApi/blob/8e6e218b2db0a868df68a33ff3952dd262f000cc/FrontEnd/styleguide/css/styleguide.css#L16
- https://github.com/opentable/BookARestaurant_MVC3/blob/49bbb18d1f89b3e31feca0c10eaf3b7fb87a24e0/CMS/Cms_Data/Contents/CompetitionDB/Media/Assets/CSS/holding/Holding.css#L131
- https://github.com/opentable/CHARM/blob/9b567de0963fd81670772508a234837cdc884540/FrontEnd/styles/charm_3.css#L6
- https://github.com/opentable/external-api-developer-portal/blob/d899e6b2784dea4b8496ae05a43ffe13290ca9f2/public/css/master.css#L19
- https://github.com/opentable/BookARestaurant_MVC3/blob/49bbb18d1f89b3e31feca0c10eaf3b7fb87a24e0/BAR.Web/Assets/CSS/admin-error.css#L69
- https://github.com/opentable/BookARestaurant_MVC3/blob/49bbb18d1f89b3e31feca0c10eaf3b7fb87a24e0/CMS/Cms_Data/Contents/CompetitionDB/Media/Assets/CSS/admin-error.css#L69
- https://github.com/opentable/BookARestaurant_MVC3/blob/49bbb18d1f89b3e31feca0c10eaf3b7fb87a24e0/CMS/Cms_Data/Contents/CompetitionDB/Media/Assets/CSS/AdminSite.css#L35
- https://github.com/opentable/tracker-srvc/blob/6ef2898d6fe34aa7fee0630bc6a99e332c6e95a4/public/js/bower_components/angularjs-dropdown-multiselect/pages/stylesheets/print.css#L17
- https://github.com/opentable/BookARestaurant_MVC3/blob/49bbb18d1f89b3e31feca0c10eaf3b7fb87a24e0/BAR.Web/DTDManage/Config/Telerik/EditorContentAreaStyles.css#L120
- ## https://github.com/opentable/neighborhood-lookup-api/blob/1b5b83e958156cdb4f16e9221dd60f7a91adb30e/src/OpenTable.NeighborhoodLookup.API/ClientApp/src/components/NavMenu.css#L8
- https://github.com/opentable/BookARestaurant_MVC3/blob/49bbb18d1f89b3e31feca0c10eaf3b7fb87a24e0/CMS/Cms_Data/Sites/bar_help/Themes/Default/Site.css#L34
- There is just a lot in CRM, so just assume that entire REPO does that 100% thing
- https://github.com/opentable/waitlist-ios/blob/d370a59f59ec3348a211ea7499a4d63aa6834fc1/Guides/Best%20Practices.playground/Documentation/stylesheet.css#L2
- https://github.com/opentable/LegacyApi/blob/8e6e218b2db0a868df68a33ff3952dd262f000cc/FrontEnd/cobrands/lastminute.css#L2
- https://github.com/opentable/synchronizer-ui/blob/5a9f1194e154fdcea3337b041d32218b46f29d7f/src/components/layout.css#L658
- https://github.com/opentable/attic/blob/55be5f8fc2acee6acb2a64a7a931da83d077e93c/larder/src/larder.web/web/styles.css#L2

Any more...

ronniechong:
I am for the `rem` direction and I am also wary the need to retain the `pixels` tokens if some engineers want to keep using it. Should we have a fallback token if they do not wish to use `rem`? e.g `token.yml` has rem, `token-pixel.yml` has pixels as legacy. Can't think of a better label, `-fallback`, `-legacy`, `-pixel`.

That means, the otkit-spacing npm package will then:
```
import { spacingXSmall } from 'otkitcolors/token'; // imports 0.25rem
import { spacingXSmall } from 'otkitcolors/token-pixel'; // imports 4px
```

dethstrobe:
I think we should actually not leave the legacy pixel measurement. I think we should go all in or else there might be some temptation to not move forward with new standards. While there might be some weird edge case breaks, I think it'd still be worth the risk and to find those edge cases and fix them would be more important then supporting legacy standards.

ronniechong:
> I think we should actually not leave the legacy pixel measurement. I think we should go all in or else there might be some temptation to not move forward with new standards. While there might be some weird edge case breaks, I think it'd still be worth the risk and to find those edge cases and fix them would be more important then supporting legacy standards.

I think Consumer side uses these packages heavily so we need to check with their side if deprecating pixels completely the way to go

char1e5 (approved):
@dethstrobe Thanks to reach out on this PR. From consumer side, we definitely use `rem` on the page. 
Earlier consideration for design token has broader use cases, like app, presenting more universal unit value and legacy standards. Time changed, I don't think we should be blocked by this.

andreiduca:
I'm in favor of moving to rem values instead of pixel, BUT... what is the greater goal here? and how will he make sure changing this now will be 100% backwards compatible? what happens with code (components, apps) that upgrades to these new tokens, while still using px values for non-otkit sizing?

i would like a bit more clarity into this upgrade and a concrete proposal and roadmap or migration/upgrade plan.

dethstrobe:
# Goal

Make opentable.com more responsive.

By moving to rems and ems we will be able to scale up or down the content quickly based off of screen size or user preference.

This will also allow us another benefit of making our site more accessible.

## Risk

Low. As long as the base font-size is 16px, this will have no effect. For any handful of app that use design tokens and change the base font, the content will scale appropriately, which should have a small effect on layout.

## Testing

- Update the values of the design tokens in the node_modules on a local copy of buffet and build.
- Link this buffet to a microapp
- Update the values of the design tokens in the node_modules the microapp to test
- run locally and observe differences.

ronniechong (approved):

neha-ot (approved):

PR #543, opened 2022-11-30
[OTK-143] update font weight for xxsmallmedium

justinelam:
## Description
Realized that I had used the wrong font-weight value `medium` -- it should be `500`

Bumped patch version.


justinelam:
@karuto would you mind reviewing and merging this PR?

